PR #1, opened 2019-12-04
Update README.md

atheurer:


README.md
```diff
@@ -1,2 +1,39 @@
...
+### Input
+
+Rickshaw needs the following:
+- Benchmark config file
+  - Rickshaw really does not care what the benchmark is; however, rickshaw must have a config file to know how to run the benchmark.  These config files are not bundled with the rickshaw project, but some from other benchmark-helper projects, or you write this config file yourself for your own benchmark.  The config file instructs rickshaw on what to execute on the controller (the host where you are running rickshaw as well as what to run on any benchmark clients and servers. 
+- Benchmark parameters
+  - This may be an array of multiple parameters, when the user wants the benchmark run mulitple ways, for example:
+    <pre>[ "--rw=read --bs=4k", "--rw=write --bs=4k" ]</pre>
+  - The [multiplex](https://github.com/perftool-incubator/multiplex) project can be used to generate this array (it can convert things like "--rw=read,write --bs=4k" into the array above), and it will do parameter validation for you as well.
+- Endpoint extension
```
k-rister: While I know what you are talking about, it probably would be helpful to include a definition of what an endpoint is.  I have a sneaking suspicion that many wouldn't understand that.

README.md
```diff
@@ -1,2 +1,39 @@
...
+  - This determines how the benchmark gets executed on different endpoints.  The default extension, local, simply runs the benchmark command on the local host.  Rickshaw supports using multiple extensions for different endpoints.  For example, if you want to run uperf benchmark, you need both a client and server uperf.  If you want to run the uperf server on Kuberbetes, but you want to run the uperf client on a baremetal host, you can use the 'k8s' extension for the server and the 'ssh' extension for the client
+  - <pre>--endpoint:server[1]:k8s:$master-hostname --endpoint:client[1]:ssh:$client-hostname</pre>
+  - Other examples
+    - 8 servers running in 8 containers in k8s and 8 clients running on the same barmetal host:  
+      - <pre>--endpoint:server[1-8]:k8s:$master-hostname --endpoint:client[1-8]:ssh:$client-hostname</pre>
+    - 8 servers running in 8 containers in k8s cluster A and 8 clients running in 8 containers in k8s cluster B:  
+      - <pre>--endpoint:server[1-8]:k8s:$cluster-a-master-hostname --endpoint:client[1-8]:k8s:$cluster-b-master-hostname</pre>
```
k-rister: I like this endpoint definition format.  You should probably also include examples for non master/server workloads as I'm not sure what that would look like.  For example, for something like the hypothetical ddbench which is a single instance benchmark would it be something like these:

```
--endpoint:local
--endpoint:ssh:$hostname
```

And potentially more crucially, what if you want to run multiple independent copies?

```
--endpoint:[1-8]:k8s:$cluster-master-hostname
```

README.md
```diff
@@ -1,2 +1,39 @@
...
+    <pre>[ "--rw=read --bs=4k", "--rw=write --bs=4k" ]</pre>
+  - The [multiplex](https://github.com/perftool-incubator/multiplex) project can be used to generate this array (it can convert things like "--rw=read,write --bs=4k" into the array above), and it will do parameter validation for you as well.
+- Endpoint extension
+  - This determines how the benchmark gets executed on different endpoints.  The default extension, local, simply runs the benchmark command on the local host.  Rickshaw supports using multiple extensions for different endpoints.  For example, if you want to run uperf benchmark, you need both a client and server uperf.  If you want to run the uperf server on Kuberbetes, but you want to run the uperf client on a baremetal host, you can use the 'k8s' extension for the server and the 'ssh' extension for the client
+  - <pre>--endpoint:server[1]:k8s:$master-hostname --endpoint:client[1]:ssh:$client-hostname</pre>
+  - Other examples
+    - 8 servers running in 8 containers in k8s and 8 clients running on the same barmetal host:  
+      - <pre>--endpoint:server[1-8]:k8s:$master-hostname --endpoint:client[1-8]:ssh:$client-hostname</pre>
```
k-rister: Should this be:

```
--endpoint:server[1-8]:k8s:$cluster-master-hostname --endpoint:client[1-8]:ssh:$client-hostname
```

README.md
```diff
@@ -1,2 +1,39 @@
...
+    - 8 servers running in 8 containers in k8s cluster A and 8 clients running in 8 containers in k8s cluster B:  
+      - <pre>--endpoint:server[1-8]:k8s:$cluster-a-master-hostname --endpoint:client[1-8]:k8s:$cluster-b-master-hostname</pre>
+  - If you don't provide any endpoint options, rickshaw will assume you want to run on the local system, and only 1 client will be used, and if the benchmark requires a server, only 1 server will be used, also on the local host.  
+- User and test information
+  - The following is optional, but is highly recommended to use so that your run can be easily searchable later.  Rickshaw looks for the following
+    - RS_NAME Environment variable or --name option with user's full name "First Last"
+    - RS_EMAIL Environment variable or --email option with user's email address "my-email@email.domain"
+    - RS_TAGS Environment variable or --tags containing a comma-separated list of words that are relevant to the run
+    - RS_DESC Environment variable or --desc containing a free form description of the purpose, conditions, or any other relevant information about this test.
+
```
k-rister: Just something to think about, I would like the ability to edit these after a run is complete and before the result is sent somewhere else.  I'm pretty sure we have discussed having the UI allowing editing of this information, especially tags, but I also think it would be a good idea to let the user do that on the system.

README.md
```diff
@@ -1,2 +1,39 @@
...
+    - RS_EMAIL Environment variable or --email option with user's email address "my-email@email.domain"
+    - RS_TAGS Environment variable or --tags containing a comma-separated list of words that are relevant to the run
+    - RS_DESC Environment variable or --desc containing a free form description of the purpose, conditions, or any other relevant information about this test.
+    
+  
+### Output
+  
+Rickshaw will provide the following:
+- Human readable log of the rickshaw execution, including any warning and errors from other projects, tools, or benchmarks it used.
+- A JSON file or Elastic documents describing the benchmark run that was handled by rickshaw
```
k-rister: I had not really considered both a JSON file AND the Elastic CDM documents as being output but I suppose that makes sense.  Would the JSON file be an intermediary format used to create the CDM documents?

README.md
```diff
@@ -1,2 +1,39 @@
...
+- Raw output from tools and benchmarks used.
+- Post-processed data (optionally) from tools and benchmarks
+  - Tools and benchmarks run by rickshaw should have post-processing programs to convert its native output to [CommonDataModel](https://github.com/perftool-incubator/CommonDataModel)
+  - Post-processing this data can be done (and re-done) at a later time if the user chooses.
```
k-rister: I think we need to be mindful of what information needs to be packaged with the data to allow for delayed/offsite postprocessing.  Most tools are really simple, but something like perf concerns me.

README.md
atheurer: For the ddbench example, I think we'll just clarify that benchmarks either are client or client-server, and some have just 1  while others can have many client/servers.  Now, perhaps we can have a shortcut where the user does not specify client[] or server, we assume they mean client and there's only 1 client for the benchmark.

for <pre> --endpoint:[1-8]:k8s:$cluster-master-hostname </pre>

So this is for multiple invocations of the endpoint extension?  I am not sure how I feel about that, but I guess it could be possible.  I suppose for a k8s extension, each invocation of that extension is a different pod.  I like the idea of multiple invocations/pods, but I am worried using a single --endpoint argument might be too difficult, because we still may need to use "client[n-m]" and "server[n-m]".  Here are a few examples I can think of:

1 pod on 1 k8s cluster, 8 clients and 8 servers come from the same pod:
<pre>--endpoint:k8s:[1-8]:$cluster-master-hostname</pre>

8 pods on 1 k8s cluster, 1 client and 1 server per pod:
<pre>
--endpoint:k8s:[1]:$cluster-master-hostname
--endpoint:k8s:[2]:$cluster-master-hostname
--endpoint:k8s:[3]:$cluster-master-hostname
--endpoint:k8s:[4]:$cluster-master-hostname
--endpoint:k8s:[5]:$cluster-master-hostname
--endpoint:k8s:[6]:$cluster-master-hostname
--endpoint:k8s:[7]:$cluster-master-hostname
--endpoint:k8s:[8]:$cluster-master-hostname
</pre>

2 pods on 1 k8s cluster, 8 clients for 1 pod and 8 servers for another pod:
<pre>
--endpoint:k8s:client[1-8]:$cluster-master-hostname
--endpoint:k8s:server[1-8]:$cluster-master-hostname
</pre>

I suppose this can work, but you are limited to the same client/server[] description for all of the pod invocations:
<pre>
--endpoint[2]:k8s:[1-8]:$cluster-master-hostname
</pre>

Note that the [2] above is a quantity and not an ID.  The above argument implies that each pod has both the client and the server, and there are a total of 2 pods, 16 clients, and 16 servers.

README.md
atheurer: actually I think I should have used: 
<pre> --endpoint:k8s:server[1-8]:$cluster-master-hostname --endpoint:ssh:client[1-8]:$client-hostname </pre>

So the format would be:  --endpoint:<extension>:$type[$range], where $type is client, server or blank, and $range is n-m or if blank, assumed to be the first client/server.

README.md
atheurer: I think we need some ability to re-import the CDM data, so you can make your edits, either via changing files or running some kind of 'crucible run' command to update, the run 'crucible import --update' to get the ES docs updated and not duplicated.

README.md
atheurer: I was just thinking that users trying to parse multiple CDM docs might be difficult, and so having a JSON doc which at least captures the run data (run/iteration/sample/period docs, but not the metric docs) would be nice.  I think we can make it so we use this JSON doc to generate the ES docs.  We could even go straight from this JSON doc to importing into ES (many docs) and skip the local-generation of many CDM docs.

It would also be a lot easier IMO to change something like the run.description in the JSON by hand if you wanted to.

There's a possibility we could do the same for metrics, but I would table that until we are certain it works well for the run/iteration/sample/period docs.

README.md
atheurer: For any of these tools we just need to make sure we capture all of the necessary data to re-postprocess, including using the native tool's reporting utility.  For perf the archive/restore thing should be adequate.  And while were on the topic, I would like perf to use that interval thing where a new perf.data file can be generated for something like 1/second.  That way when we eventually do write a visualizer for perf, we can change the time range of what we are seeing a little easier.

README.md
k-rister: It's not clear to me which of the above you prefer, but I think you are on the right track.  I want to be clear I wasn't necessarily advocating for the one I suggested, it just seemed somewhat appropriate given the lack of a client/server label.  I do wonder if it should just be very explicit and hierarchical:

```
--endpoint:k8s:pods[1-2]:client[1-8]:$cluster-master-hostname
--endpoint:k8s:pods[3-4]:server[1-8]:$cluster-master-hostname
```
For 2 pods with 8 clients each and 2 more pods with 8 servers each.

If you just want to refer to everything as a client and add in the server case where needed I'm fine with that, I just think it should be documented if that is the case.

README.md
k-rister: As I suggested above, how about:

```
--endpoint:${endpoint-type}:${endpoint-subtype}[$range]:{$type[$range]}

Where ${endpoint-type} could be one of "local", "ssh", "k8s", etc.

Where ${endpoint-subtype} would be context sensitive for the ${endpoint-type}, for "k8s" it could be "pods" otherwise assumed to be all in one pod

The rest as you said.

```

README.md
k-rister: That would be nice, one of my biggest gripes with the current pbench stuff is not being able to easily replace uploaded data with something that has been updated.

I think I would prefer to allow the user to run some crucible commands to update the information rather than forcing them to manually edit files.  Something like:

```
crucible list-run-ids
<list of runs with ids that are available locally>

crucible run [--run-id=<id>] get-tags
<list of tags dumped>

crucible run [--run-id=<id> set-tags "<list of tags>"
```

README.md
k-rister: Yeah, I'm pretty keen to see a perf implementation work in the way your describe.

README.md
atheurer: The problem here is that the client[1-8] does not just mean 8 clients, but clients IDs 1-8, and so you can't have client IDs 1-8 for pod1 and pod2 (and same thing for servers).  Assigning this ID specifically is important because we want to mix this with other endpoints like local or ssh, and we need to connect the dots between specific clients and servers with the same ID.  And so while it may require more --endpoint options, we may have to do something like:

<pre>
--endpoint:k8s:client[1-8]:$cluster-master-hostname
--endpoint:k8s:client[9-16]:$cluster-master-hostname
--endpoint:k8s:server[1-8]:$cluster-master-hostname
--endpoint:k8s:server[2-16]:$cluster-master-hostname
</pre>

That is, unless you want to have a format like:

<pre>
--endpoint:k8s:pod[1]:client[1-8],pod[2]:client[2-16],pod[3]:server[1-8],pod[4]:server[9-16]:$cluster-master-hostname</pre>

Actually I think this may be a good time to just assert that the format is:

<pre> --endpoint:$endpoint_type:$endpoint_options</pre>

And the $endpoint_options format is specific to that endpoint.  And it is the responsibility for that endpoint type to provide a way to identify clients and servers.

k-rister: Good point, different endpoints will/could have vastly different options so trying to standardize across all of them it probably a futile effort.

For the k8s endpoint I like your last suggestion:

```
--endpoint:k8s:pod[1]:client[1-8],pod[2]:client[2-16],pod[3]:server[1-8],pod[4]:server[9-16]:$cluster-master-hostname
```

PR #2, opened 2020-01-15
Atheurer patch 1

atheurer:
Initial rickshaw script with just the overall layout.  This does not actually run a workload yet.  The 'local' endpoint exists, but does not do anything.  Cmdline param verification works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # rickshaw
-Rickshaw will run a benhcmark for you.  It "takes" your benchmark wherever you need it to go, as long as there is an implementation for your particular endpoint (a host, cloud, container-runtime, etc).  These endpoints are in ./endpoints, and as of this version, only the 'local' endpoint exists.  
+Rickshaw will run a benchmark for you.  It "takes" your benchmark wherever you need it to go, as long as there is an implementation for your particular endpoint (a host, cloud, container-runtime, etc).  These endpoints are in ./endpoints, and as of this version, only the 'local' endpoint exists.
 
 ### Input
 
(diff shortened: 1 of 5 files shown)


rickshaw
```diff
@@ -0,0 +1,212 @@
...
+}
+# Confirm we have required params
+defined $bench_config_ref || die "[ERROR]You must use --bench-config=/path/to/benchmark-config.json\n";
+defined $params_ref || die "[ERROR]Your must use --bench-params=/path/to/benchmark-params.json\n";
+
+-e $base_dir || mkdir($base_dir);
+
+# If there are no endpoints, assume 1 endpoint using the 'local' extension
+if (scalar @endpoints == 0) {
+    if (exists $bench_config{'client'} and exists $bench_config{'server'}) {
```
k-rister: This might be a stupid question, but would you ever use a server without a client?

rickshaw
```diff
@@ -0,0 +1,212 @@
...
+
+print "\nEndpoints that will be used:\n";
+foreach my $endpoint (@endpoints) {
+    printf "%s\n", $endpoint;
+}
+$run{'endpoints'} = \@endpoints;
+
+
+printf "\nendpoint output:\n";
+# Deploy the endpoints so they are ready to run benchmark and tools
```
k-rister: So is endpoint deployment where packages would get installed, containers/vms launched, etc.?

README.md
```diff
@@ -32,6 +31,17 @@ Rickshaw needs the following:
     - RS_EMAIL Environment variable or --email option with user's email address "my-email@email.domain"
     - RS_TAGS Environment variable or --tags containing a comma-separated list of words that are relevant to the run
     - RS_DESC Environment variable or --desc containing a free form description of the purpose, conditions, or any other relevant information about this test.
+
+Other, optional paramers include:
```
k-rister: What about the mode of running different iterations and/or samples on different nodes?

rickshaw
```diff
@@ -0,0 +1,212 @@
...
+        for (my $iid = 0; $iid < scalar @params; $iid++) {
+            my %test = ('iter_id' => $iid, 'samp_id' => $sid);
+            push(@tests, \%test);
+        }
+    }
+}
+
+# Cycle through the list of tests using roadblock to kick them off
+my @iterations;
+printf "\nRequesting endpoints to start tools\n";
```
k-rister: Looks like I have some work to do....

testing/fio-user-params.json
```diff
@@ -0,0 +1,23 @@
+[
```
k-rister: so is this an example of the output from multiplex?

rickshaw
```diff
@@ -0,0 +1,212 @@
...
+my @params;
+my @endpoints;
+my %run; # A multi-dimensional, nested hash, schema TBD
+         # This hash documents what was run.
+         # CDM utilities will convert this into many CDM documents.
+chomp(my $base_dir = `/bin/pwd`);
+
+sub debug_log {
+    if ($debug) {
+        print "[DEBUG]" . $_[0];
```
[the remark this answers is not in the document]
atheurer: I tried that but it did not work.  I think any sub gets just the @_ array.

rickshaw
atheurer: I don't think so.

rickshaw
atheurer: Yes, and then they need to be on "stand-by", issuing their first in a list of commands, which would likely be a roadblock command signaling that they are ready to run.  The rest of the commands would be tool start/stops, roadblock commands, and benchmark executions.  I think we'll need to pass either a script or a "queue of commands" when we build the endpoint filesystem-image.  Rickshaw will be responsible for building these commands for each client & server.  I thought about having each endpoint do that, but I would like to avoid that because I don't want to duplicate that work in many endpoints.  Whatever the endpoint may be (local, ssh, k8s, osp-vm, etc), the filesystem-image should be consistent, and therefore the commands it issue should be consistent regardless of the endpoint type.  The only real difference in the commands are things like IP addrs and ports.

README.md
atheurer: Are you referring to the parallel execution idea?  I have definitely not put much thought into that yet.  If we did support that, I wonder if it should be a different endpoint type, like a special version of "ssh" endpoint.  I don't know...  Not a real concern at this point.

testing/fio-user-params.json
atheurer: I don't know if I actually have multiplex doing this yet, but it will.

rickshaw
atheurer: Well, we will need roadblock client included in the workshop images for sure.  For now for "local" endpoint I can probably just manually install all the redis client stuff (and some benchmark) and concentrate on the other stuff for the local endpoint, like processing the script or command-queue and making sure roadblock communication works.  Then hopefully most of that can be re-used for the ssh endpoint, and when workshop is ready, we'll switch to the fs-image roadblock provides (with chroot or something).

README.md
k-rister: I think so.  I guess I thought it would be similar in concept to your current idea of i-s vs. s-i.  I would have thought it would be endpoint agnostic, but maybe I don't fully grasp the implications.

k-rister:
Did you mean to merge this into atheurer-patch-1 branch of the upstream
repo?

--
Karl Rister <krister@redhat.com>

atheurer:
Not really, but I merged an older outstanding PR which seems to have gotten this one into master.

PR #15, opened 2020-04-06
Add more tool functionality including blacklist

atheurer:
-blacklist in a tool's config ensures tools are not run
 on a client/server (the default location)
-whitelist in a tool's config ensures tools are run somehwere
 other than a client/server (a collector, of which the
 endpoint is responsible to running)
-tool data is organized in per-tool directories
-tool commands are broekn out by collector and cmd files
 have tool labels (sysstat: <start-cmd>)
-when specifying a container image all of "endpoint-run" is not
 copied, and only "client-server-script" is copied
-use $cs_label instead of $cs_id to mean "<client/server>->id>"
-update the tool config schema to recognize the k8s
 endpoint (instead of rhocp).


endpoints/k8s/k8s
```diff
@@ -383,7 +383,7 @@ while read line; do
             exit_error "Exiting: exit code 2 (timeout) from roadblock iteration $iter, sample $samp"
         fi
     done
-done < "$rickshaw_run_dir/client-server-bench-commands/client-1"
+done < "$rickshaw_run_dir/bench-cmds/client/1"
```
k-rister, line 386: The hardcoding of "1" seems curious here.  I guess I missed this in the previous set of K8S patches.  Why is it hardcoded?

endpoints/local/local
```diff
@@ -348,7 +350,7 @@ while read line; do
             exit_error "Exiting: exit code 2 (timeout) from roadblock iteration $iter, sample $samp"
         fi
     done
-done < "$rickshaw_run_dir/client-server-bench-commands/client-1"
+done < "$rickshaw_run_dir/bench-cmds/client/1"
```
k-rister, line 353: Again, "1" hardcoded.

client-server-script
```diff
@@ -27,7 +27,7 @@ function do_roadblock() {
     local cmd=""
     cmd="$cmd /usr/bin/roadblock.py --role=follower --redis-server=$roadblock_server"
     cmd="$cmd --timeout=$timeout --redis-password=$roadblock_passwd"
-    cmd="$cmd --uuid=$roadblock_id:$label --follower-id=$cs_id $extra"
+    cmd="$cmd --uuid=$roadblock_id:$label --follower-id=$cs_label $extra"
```
k-rister, line 30: I don't care one way or the other, but I'm just curious what the reasoning behind this change in variable name was (especially since it's so pervasive in this PR).  Is this indicative of some issue you were trying to resolve?

atheurer: I am just trying to avoid confusion as to what cs_id means vs cs_label.  An "id" is  just the number while label is the $type-$id, like client-1.  When first writing this I did not need to use just the id (1, 2..) but now I do in some places.

endpoints/k8s/k8s
atheurer: It is hardcoded as 1 because it does not matter which client's file it uses as long as the file is there.  Client "1" is guaranteed to be there.

endpoints/local/local
atheurer: same as above

client-server-script
k-rister: ack

endpoints/k8s/k8s
k-rister: ack, makes sense

endpoints/local/local
k-rister: ditto

PR #17, opened 2020-04-20
Require bench server both start and stop

atheurer:
-A bit more reliable than assuming server start script creates a pid file
-Also fix for RS_CS_LABEL assignment
-Also come debug code for running workshop


atheurer:
There are some minor fixes and info output here and there

Issue #71, opened 2020-08-24
Don't start tool pods for masters if masters are not detected


atheurer:
Should seem pretty obvious, but k8s is trying to start pods for an empty list of masters and fails.  Don't do this.

k-rister:
Why is the masters list empty?

atheurer:
See https://github.com/perftool-incubator/rickshaw/issues/70

atheurer:
Fixed with PR #72

Issue #70, opened 2020-08-24
Use -o json when using 'kubectl get' to get more reliable output


atheurer:
Don't parse human-readable output when we could be using JSON.  Currently 'grep' is used and is prone to errors when the human readable output changes or grep does not account for all possible output.

atheurer:
We already did this for finding workers so reuse for masters.  Fixed with PR #72

PR #72, opened 2020-08-25
k8s: warn on no master/worker tool pods, get masters more reliably

atheurer:
-use same logic we have for getting workers (using json and jq)


atheurer:
fixes issue #71

atheurer:
Also fixes issue #70

PR #82, opened 2020-10-30
rickshaw-run: bug fixes in dump_params

k-rister:
- The output from multiplex.go produces 'value' and not 'val'.

- Detect whether a parameter has a value or not and act accordingly.

### DIFF
--- a/rickshaw-post-process-bench
+++ b/rickshaw-post-process-bench
@@ -50,10 +50,14 @@ sub dump_params {
     foreach my $param (@{ $params_ref }) {
         my $arg = $$param{'arg'};
         my $val = $$param{'val'};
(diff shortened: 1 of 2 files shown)


k-rister:
@atheurer 

I'm not sure whether you want to move to 'value' or stay with 'val' -- either way something needs to be fixed.

The other changes fix the spewing of messages that I was seeing.

atheurer:
This will probably throw off a lot of existing bench-params.json files out there, but I suppose that is not a big deal if you think 'value' is more clear.

k-rister:
I really don't care, I will defer to you since most of this is your brain child.

atheurer:
let's keep val and change multiplex

PR #84, opened 2020-11-16
rickshaw-run: provide simple json validation of the bench-params.json…

k-rister:
… input


k-rister:
While working on multiplex I accidentally created a bad bench-params.json and submitted it to rickshaw.  That resulted in all kinds of nastiness so it would be nice to have simple JSON validation that ensure that the file is properly formatted and has the correct properties present.

atheurer (approved):

PR #92, opened 2020-12-08
handle xz compressed files

k-rister:


atheurer (approved):
This looks fine, but I wonder if we might need to compress written files after they are written uncompressed, because we could do this with many threads (-T 0) with xz.  This might only be necessary if the files are really big, and the transparent compression by perl is too slow.

k-rister:
Yeah, I considered that.  That would require temporarily having the necessary disk space available.  I believe I wrote it so that the output type is determined by the name used, so if we find this to be too slow we just have to change the name back and add in the other method.

PR #101, opened 2021-01-05
Ensure endpoints and client/servers access the controller via IP

atheurer:
-Providing the host to these via 'hostname' is not good enough.
-The hostname can be wrong, or not resolved to an IP by the
 clients/servers.
-This does require that the crucible-controller have bind-utils
 and iproute2 packages


atheurer:
In this commit, the roadblock server is assumed to be the same as the rickshaw_host (also referred to as the controller_ipaddr).  There is no reason to make these different, since the redis server is always on the user's crucible controller.

PR #106, opened 2021-02-01
A number of changes for uperf and trafficgen

atheurer:
-ensure messages are processed by specific endpoint
-add max-sample-failures as option
-add support for kata


bootstrap
```diff
@@ -130,6 +130,12 @@ while true; do
             export disable_tools="$1"
             shift;
             ;;
+        --max-sample-failures)
+            shift;
+            echo "[WARNING] --max-sample-failures was not used, so seeting to default of $max_sample_failures"
```
k-rister, line 135: This seems wrong...

atheurer: yeah I forgot to change this.  I will in the next merge

PR #112, opened 2021-02-25
Manage userenvs within rickshaw

atheurer:
-workshop would not track userenvs, only use them
-any project using workshop should manage its own userenvs
-included in this commit is a modular approach to build userenvs
 -use build.sh to regenerate userenv jsons

### DIFF
--- a/rickshaw-run
+++ b/rickshaw-run
@@ -353,7 +353,7 @@ sub build_container_image {
         my $workshop_cmd = $run{'workshop-dir'} . "/workshop.pl" .
             # " --skip-update=true " .
             " --config " . $cs_conf_file .
-            " --userenv " . $run{'workshop-dir'} . "/userenvs/" . $userenv . ".json" .
+            " --userenv " . $rickshaw_project_dir . "/userenvs/" . $userenv . ".json" .
             " --requirements " . $run{'bench-dir'} . "/workshop.json" .
             " --requirements " . $cs_req_file .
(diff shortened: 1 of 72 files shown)


userenvs/build.sh
```diff
@@ -0,0 +1,36 @@
...
+
+userenv="$1"
+if [ ! -d "$userenv" ]; then
+    echo "Directory $userenv not found"
+    exit 1
+fi
+
+
+pushd "$userenv"
+echo "{" >../staging.json
```
k-rister, line 16: I'm just curious, does `jq` allow you to manipulate/create JSON docs easier than manually doing all these `echo` commands?

rickshaw-run
```diff
@@ -353,7 +353,7 @@ sub build_container_image {
         my $workshop_cmd = $run{'workshop-dir'} . "/workshop.pl" .
             # " --skip-update=true " .
             " --config " . $cs_conf_file .
-            " --userenv " . $run{'workshop-dir'} . "/userenvs/" . $userenv . ".json" .
+            " --userenv " . $rickshaw_project_dir . "/userenvs/" . $userenv . ".json" .
```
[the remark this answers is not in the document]
atheurer: yes or config/workshop/userenvs

userenvs/build.sh
atheurer: I think it is strictly query, but there may be a program for that.

PR #119, opened 2021-03-09
Add support for the cpu-partitioning endpoint parameter

k-rister:


k-rister:
I can't tell you how many times I did a typo on "partitioning" while writing this...made it take forever...

k-rister:
So far this is just the plumbing through rickshaw and the endpoints to get the user request to the bootstrap script that will eventually do most of the heavy lifting (that is the next step...).

atheurer (approved):
LGTM

rafaelfolco:
lgtm

rafaelfolco (approved):

bootstrap
```diff
@@ -168,6 +168,21 @@ if [ ! -e $ssh_id_file ]; then
...
 export config_dir="$base_run_dir/config"
 export client_server_config_dir="$config_dir/client-server"
+
+echo "cpu-partitioning:"
+if [ "${cpu_partitioning}" == "1" ]; then
+    echo "Enabled"
+
+    environment="process"
+    if [ "${endpoint}" == "k8s" ]; then
+	environment="container"
```
atheurer, line 178: We don't need to worry about it now, but eventually we need to figure out what to do for kata and cnv.  endpoint will be "k8s", but osruntime will be kata or cnv.  However, this problem is bigger than just us, as we need the VM to be configured in such a way to provide isolation (unless we find a way to do it 100% at runtime).

discover-cpu-partitioning.py
```diff
@@ -0,0 +1,173 @@
...
+        print("all cpus: %s" % (system_cpu_topology.formatted_cpu_list(all_cpus)))
+        print("all cpus: %d" % (len(all_cpus)))
+        print("")
+        online_cpus = system_cpus.get_online_cpus()
+        print("online cpus: %s" % (online_cpus))
+        print("online cpus: %s" % (system_cpu_topology.formatted_cpu_list(online_cpus)))
+        print("online cpus: %d" % (len(online_cpus)))
+        print("")
+        rcu_nocbs_cpus = get_rcu_nocbs()
+        print("rcu_nocbs: %s" % (rcu_nocbs_cpus))
```
atheurer, line 112: I wonder if we should also consider (1) systemd's CPUAffinity or (2) taskset -pc 1 or taskset -pc $$, if this is not a container.  It's possible we could have lesser isolation, or user-space only isolation.

discover-cpu-partitioning.py
```diff
@@ -0,0 +1,173 @@
...
+            housekeeping_cpus.append(cpus_allowed.pop(0))
+            isolated_cpus = copy.deepcopy(cpus_allowed)
+
+        print("housekeeping cpus: %s" % (housekeeping_cpus))
+        print("housekeeping cpus: %s" % (system_cpu_topology.formatted_cpu_list(housekeeping_cpus)))
+        print("housekeeping cpus: %d" % (len(housekeeping_cpus)))
+        print("")
+        print("isolated cpus:     %s" % (isolated_cpus))
+        print("isolated cpus:     %s" % (system_cpu_topology.formatted_cpu_list(isolated_cpus)))
+        print("isolated cpus:     %d" % (len(isolated_cpus)))
```
atheurer, line 168: Is this going to set env vars, or is bootstrap going to check output and then set env vars?

atheurer (commented):
questions inline

bootstrap
k-rister: Yep.  I'm pretty much assuming that this point I'm just laying down a framework that will cover a few cases and we will have to expand on it as we add other endpoints/behaviors.

discover-cpu-partitioning.py
k-rister: I think so.  This is extendable and just a first pass so we can certainly add on additional methods.

discover-cpu-partitioning.py
k-rister: bootstrap is going to check the output and set the vars.  bootstrap has to do it so that children of bootstrap can see and inherit the values.  If this script did it then it would disappear as soon as this script exited.

discover-cpu-partitioning.py
```diff
@@ -122,26 +129,17 @@ def main():
...
-        print("all cpus: %d" % (len(all_cpus)))
-        print("")
+        output_cpu_info("all", all_cpus)
+
         online_cpus = system_cpus.get_online_cpus()
-        print("online cpus: %s" % (online_cpus))
-        print("online cpus: %s" % (system_cpu_topology.formatted_cpu_list(online_cpus)))
-        print("online cpus: %d" % (len(online_cpus)))
-        print("")
+        output_cpu_info("online", online_cpus)
```
rafaelfolco: this block is common to process and container envs. You might want to save a few lines of code.

discover-cpu-partitioning.py
```diff
@@ -92,27 +92,34 @@ def get_pid_cpus_allowed(pid):
...
+    if t_global.args.debug:
+        debug("%s cpus: %d" % (label, len(cpu_list)))
+        short_cpu_list = system_cpu_topology.formatted_cpu_list(cpu_list)
+        formatted_short_cpu_list = ','.join(short_cpu_list)
+        debug("%s cpus: %s" % (label, formatted_short_cpu_list))
+
+    cpu_list = ','.join(map(str, cpu_list))
+    print("%s cpus: %s" % (label, cpu_list))
+
+    print("")
```
rafaelfolco, line 105: \n saves you one lline of code

rafaelfolco (commented):
you could save a few lines of code

bootstrap
```diff
@@ -177,7 +177,20 @@ if [ "${cpu_partitioning}" == "1" ]; then
     if [ "${endpoint}" == "k8s" ]; then
 	environment="container"
     fi
-    /usr/local/bin/discover-cpu-partitioning.py --environment=${environment}
+    cpu_data=$(/usr/local/bin/discover-cpu-partitioning.py --environment=${environment})
+
+    # log the data returned
+    echo -e "${cpu_data}"
```
rafaelfolco, line 183: print on debug mode only ?

bootstrap
```diff
@@ -177,7 +177,20 @@ if [ "${cpu_partitioning}" == "1" ]; then
...
+    export HK_CPUS=${housekeeping_cpus}
+
+    isolated_cpus=$(echo -e "${cpu_data}" | grep isolated | awk '{ print $3 }')
+    export WORKLOAD_CPUS=${isolated_cpus}
+
+    # pin myself to the housekeeping cpus so that I and my children
+    # are contained there
+    taskset --cpu-list --pid ${HK_CPUS} $$
 else
     echo "Disabled"
```
rafaelfolco, line 195: "cpu partitioning disabled" ? also on debug mode only ?

discover-cpu-partitioning.py
k-rister: Yeah, I was waiting until I'm absolutely sure it is going to stay this way before consolidating...

discover-cpu-partitioning.py
k-rister: I tend to write my code  in preparation for what might happen...as in if I add a line in between the current line and the newline I would have to change the current line if I used \n, by making it a separate line I don't have to touch the current line again...

bootstrap
k-rister: This output is going into what is already an extremely verbose logfile where we can go to get information about a run if needed, suppressing output here would go against the way the rest of the script is written.

bootstrap
k-rister: Same as above.

This ends up looking something like this:

discover-cpu-partitioning.py
atheurer: OK, sounds good.  I assume bootstrap will also taskset all pids in the container (if this is a container) or any child pids (if on baremetal) to the hk_cpus before running client-server-script.

k-rister: bootstrap tasksets itself so that anything it forks will inherit the new cpu mask:

https://github.com/perftool-incubator/rickshaw/blob/6facf2fbcdb4155119c8e09514a4428c46f53267/bootstrap#L193

PR #129, opened 2021-03-26
Update roadblock output handling

k-rister:
- Recent changes in roadblock require updates to properly handle

- Properly format the printing of roadblock output so that it ends
  with a newline in shell scripts


k-rister:
This will need to be merged simultaneously to the roadblock PR that is currently open.

PR #130, opened 2021-03-30
roadblock timeout fixup

k-rister:


k-rister:
@atheurer I don't really want to merge this because it is such a hack...but this is a necessary complement to the other hack you merged before you went on PTO...

atheurer:
Does this actually fix a problem for you?

k-rister:
Yes, without it the problem that you tried to fix with this https://github.com/perftool-incubator/rickshaw/commit/13dff3674a0d882b26416232d37f9a2f82ab23ad still exists.

atheurer:
what kind of benchmark do you see this problem and how long does it typically run?

k-rister:
That was on the 12 hour cyclictest/oslat runs.

rafaelfolco:
@k-rister why don't you just change to $timeout and fallback to $default_timeout (86400) if $timeout is not set ? much better than hard coding.

atheurer (approved):

PR #152, opened 2021-05-25
Have all roadblocks lead by the controller

atheurer:
-Don't have endpoints lead roadblocks for tool-only osruntimes
-Instead, have endpoints send new follower IDs to the controller
-Then controller knows about the new folowers
-Eliminates odd, 24-hour timeouts for some roadblocks


k-rister (approved):
I was able to run a simple test so it appears to work.

PR #153, opened 2021-05-26
make cpu-partitioning a per engine setting for remotehost endpoint

k-rister:


k-rister:
@atheurer This is untested, but I think it is what you need.

atheurer (approved):

PR #154, opened 2021-05-27
Add endpoint option to override controller_ip

atheurer:
-the controller_ip is the IP address an osruntime uses to
 contact the controller.  This is normally determined automatically,
 but sometimes this is not so easy.  The user can now specify:
 "controller_ip:<ipaddr> on the endpont options to override the
 automatic determination of this IP.

### DIFF
--- a/endpoints/base
+++ b/endpoints/base
@@ -597,26 +597,10 @@ function process_final_roadblocks() {
     do_roadblock endpoint-finish "follower" $default_timeout
 }
 
-function get_controller_ip() {
-    local host=$1; shift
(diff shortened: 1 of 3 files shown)


k-rister (approved):

k-rister (changes requested):
This patch looks good, but shouldn't you also remove the second copy of the get_controller_ip function?

k-rister (commented):
So the process_final_roadblocks function looks like it was also duplicated, but the process_postbench_roadblocks was not -- can you really delete both of these?

k-rister (approved):
:-)

PR #164, opened 2021-06-28
Handle per-ID args in bench-params

atheurer:


schema/bench-params.json
```diff
@@ -26,6 +26,9 @@
             "server",
             "all"
           ]
+        },
+        "id": {
+          "type": "string"
```
k-rister, line 31: We should probably impart some requirements here, right?  As in in can't be empty and it should be a numerical value, right?

PR #173, opened 2021-07-16
A misc set of updates:

atheurer:
-better container cleanup for remotehost endpoint
-suport numa node binding for remotehost
-endpoint_remotehost_cleanup was being called twice
-update userenvs for numactl

### DIFF
--- a/endpoints/remotehost/remotehost
+++ b/endpoints/remotehost/remotehost
@@ -47,6 +47,7 @@ userenv="rhubi8"
 user="root"
 ssh_opts="-o StrictHostKeyChecking=no -o PasswordAuthentication=no"
 declare -A cpuPartitioning
+declare -A numaNode
 
 function endpoint_remotehost_test_stop() {
     echo "Running endpoint_remotehost_test_stop"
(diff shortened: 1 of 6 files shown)


endpoints/remotehost/remotehost
```diff
@@ -394,8 +473,4 @@ init_common_dirs
 remotehost_req_check
 base_req_check
 launch_osruntime
-# process_prebench_roadblocks
-# process_bench_roadblocks remotehost
-# process_postbench_roadblocks
 process_roadblocks remotehost
-endpoint_remotehost_cleanup
```
k-rister, line 401: Removing this line seems broken.  I just added this back so this might be caused by an issue with rebasing your tree (https://github.com/perftool-incubator/rickshaw/commit/eee8d3b58ff59fb144509ae5b3582b27847c05a9).

k-rister (commented):
It mostly seems ok, but given that I see what looks like an error caused by an incorrect rebase I can't help but wonder what other issues are lurking...

endpoints/remotehost/remotehost
atheurer: Removing it it not broken.  endpoint_${endpoint}_cleanup is called in process_roadblocks()

k-rister: Ah, I guessed I missed that.  Carry on.

k-rister (approved):

PR #181, opened 2021-07-24
Add max-rb-attempts to longopts

atheurer:
-will not work without this on chroot osruntime


k-rister:
oops

atheurer:
> oops

BTW getopt output is completely useless with errors.  I wonder if there's a different getopt we can use

k-rister:
> > oops
> 
> BTW getopt output is completely useless with errors. I wonder if there's a different getopt we can use

I think I read that the bash builtin doesn't support long :(

PR #190, opened 2021-08-31
rickshaw-index: ensure run doc is indexed

atheurer:
-check for run doc with timeout#

### DIFF
--- a/rickshaw-index
+++ b/rickshaw-index
@@ -340,6 +340,20 @@ sub index_metrics {
     }
 }
 
+sub run_id_in_es {
+    my $id = shift;
+    my $response = http_request("POST", "localhost:9200", "/cdm" . $cdm{'ver'} . "-run/_doc/_search/",
+                                '{"query":{"bool":{"filter":[{"term":{"run.id":"' . $id . '"}}]}},"size":1000}');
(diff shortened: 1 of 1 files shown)


rickshaw-index
```diff
@@ -340,6 +340,20 @@ sub index_metrics {
     }
 }
 
+sub run_id_in_es {
+    my $id = shift;
+    my $response = http_request("POST", "localhost:9200", "/cdm" . $cdm{'ver'} . "-run/_doc/_search/",
+                                '{"query":{"bool":{"filter":[{"term":{"run.id":"' . $id . '"}}]}},"size":1000}');
```
k-rister, line 346: I'm slightly confused now...is it `run-id` or `id`?   Which is used when?

atheurer: Going forward, run-id will not be used in rickshaw-result.json for new crucible runs.  run.id in the ES doc is not the same as run-id.  For the ES docs, there are sections like run, iteration, period, etc.  The change from run-id to id in the rickshaw-result.json is to map to the ES doc with no conversions, avoiding a "run.run-id" in the ES doc.

k-rister (approved):

PR #195, opened 2021-09-06
switch to common action for ci-environment-info

k-rister:
- also update names


.github/workflows/crucible-ci.yaml
```diff
@@ -42,6 +44,8 @@ jobs:
     steps:
     - name: Checkout rickshaw
       uses: actions/checkout@v2
+      with:
+        path: rickshaw
```
rafaelfolco, line 48: wow this is pretty cool

rafaelfolco (commented):
this is still draft ? looks pretty good, I would just squash the fixups

PR #198, opened 2021-09-13
update userenvs for perl and node module installation

tc-wilson:


userenvs/components/05-rhubi-core-perl.json
```diff
@@ -0,0 +1,9 @@
+    {
+      "name": "core-perl",
+      "type": "manual",
+      "manual_info": {
+        "commands": [
+          "curl --silent -L https://cpanmin.us | /usr/bin/perl - App::cpanminus"
+        ]
+      }
+    }
```
k-rister, line 9: this seems like a duplicate of `userenvs/components/rhubi-core-perl.json` below and the name seems weird for this location...or am I missing something?

k-rister (commented):
So right now this is just enabling the perl/node support but it doesn't actually change any packages over to using it from manual, right?

PR #201, opened 2021-09-21
try the ci mechanism for testing crucible and rickshaw updates

tc-wilson:


k-rister:
This is very old and very stale...and it was never intended to be merged anyway as I recall.

PR #202, opened 2021-09-22
add support for microk8s as a valid k8s endpoint

k-rister:
- make the check for KUBECONFIG optional

- make master nodes optional


rafaelfolco (approved):
I can't spot any clear problems. What come to mind is: where is "microk8s" definition for endpoints ? how we diff k8s and microk8s ?

rafaelfolco (approved):

PR #203, opened 2021-09-23
Add 'volumes' and 'volumeMounts' to k8s to support FlexRAN

HughNhan:
Add 'volumes' and 'volumesMounts' processing to support FlexRAN.
Prior to this and for DPDK workloads, there was an implicit volumesMount created when there was a 'hugepage' param in the CLI. This behavior is preserved, but these workloads can use the new explicit volumes and volumeMounts also.


endpoints/k8s/k8s
```diff
@@ -624,12 +644,19 @@ function build_pod_spec() {
...
-    if [ "$type" == "cs" -a "$hugepage" == "1" ]; then
-        echo '       ,"volumeMounts": [' >>$json
-        echo '          { "mountPath": "/dev/hugepages",' >>$json
-        echo '            "name": "hugepage"' >>$json
-        echo '          }' >>$json
-        echo "        ]" >>$json
+    # If user provided volumeMounts.json, use it. It not, and "hugepage", do implicitly
+    if [ "$type" == "cs" ]; then
+        set +u
+        if [ ! -z "${volumeMounts[$name]}" ]; then
```
k-rister, line 650: `! -z` should be replaced with `-n`

endpoints/k8s/k8s
```diff
@@ -661,15 +689,23 @@ function build_pod_spec() {
...
-        echo '          { "name": "hugepage",' >>$json
-        echo '            "emptyDir": {' >>$json
-        echo '              "medium": "HugePages"' >>$json
-        echo '            }' >>$json
-        echo '          }' >>$json
-        echo '        ]' >>$json
+    # If user provided volumes.json, use it. It not, and "hugepage", do implicitly
+    if [ "$type" == "cs" ]; then
+        set +u
+        if [ ! -z "${volumes[$name]}" ]; then
```
k-rister, line 695: again, `! -z` should be replaced with `-n`

k-rister (commented):
Also, lets change the first line of this commit to not be specific to flexran...its really about adding features to the k8s endpoint.  The second line seems more appropriate as a title.

endpoints/k8s/k8s
HughNhan: They are equivalent, but keeping it as-is will preserve the coding usage/style in this function. All of the other similar code blocks in this function use "! -z" construct and not "-n".

endpoints/k8s/k8s
HughNhan: Ditto aboout preserving same style (using " ! -z") in this function

k-rister (commented):
I'm fine with the `! -z` stuff since its consistent with what is already there.  I'm trying to drive what I perceive to be a better coding style going forward however...

I do want to rework the commit message to not mention flexran though -- "Initial flexran commit" doesn't mean much in the context of rickshaw.

HughNhan:
> I'm fine with the `! -z` stuff since its consistent with what is already there. I'm trying to drive what I perceive to be a better coding style going forward however...
> 
> I do want to rework the commit message to not mention flexran though -- "Initial flexran commit" doesn't mean much in the context of rickshaw.

I fixed it per your review comment.

k-rister (approved):

PR #206, opened 2021-09-27
add support for --tls-verify=true|false for container push/inspect in…

k-rister:
… rickshaw-run

### DIFF
--- a/rickshaw-run
+++ b/rickshaw-run
@@ -503,6 +503,10 @@ sub build_container_image {
                 }
             }
         }
+        if (! exists $run{'reg-tls-verify'}) {
+            $run{'reg-tls-verify'} = "true";
+        }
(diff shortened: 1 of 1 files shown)


rickshaw-run
```diff
@@ -503,6 +503,10 @@ sub build_container_image {
                 }
             }
         }
+        if (! exists $run{'reg-tls-verify'}) {
+            $run{'reg-tls-verify'} = "true";
```
rafaelfolco, line 507: enable by default?

rafaelfolco (approved):
lgtm

rickshaw-run
k-rister: yes...it is actually the default behavior if you don't specify anything.

PR #209, opened 2021-10-01
handle the case where an invalid reg-tls-verify value is encountered

k-rister:
- this can happen if the Crucible config file has not been updated to
  include a value for $CRUCIBLE_CLIENT_SERVER_TLS_VERIFY.  In that
  scenario the code assumes that a value should be present so it uses
  it, which results in a blank/empty value being passed through.

- a more formal fix is required to update the Crucible config with new
  properties but in the meantime we should properly handle what
  happens if the value is not defined (or corrupted by the user,
  etc.).


rafaelfolco (approved):
looks a clean workaround

PR #214, opened 2021-10-14
userenvs: ensure we have libcap, libzstd

atheurer:
-both rhubi's now have centos repos
-tested with tool-kernel


userenvs/opensuse.json
```diff
@@ -0,0 +1,113 @@
+{
```
k-rister, line 1: You weren't supposed to commit this one...

atheurer: I thought I also did a 'remove' for the commit to netgate this out, but maybe not.

k-rister (dismissed):

PR #225, opened 2021-12-01
rickshaw-run: ensure tool args are not bungled

atheurer:
-Also consolidate code to generate cmd
-Use ash array to store args to cmd


k-rister (approved):
whoa

PR #226, opened 2021-12-02
k8s: set default user as kni

atheurer:


k-rister:
@atheurer Do we still need this?

k-rister:
Closing due to inactivity.

PR #227, opened 2021-12-02
Incremental image build

atheurer:


rickshaw-run
```diff
@@ -362,271 +626,130 @@ sub build_container_image {
...
-            }
+    }
+    $image_url = $run{'reg-repo'} . ":" . $workshop_args[$i]{'tag'};
+    # After finding the most complete image, build any "more" complete images until "most"
+    # complete is built.
+    $i++;
+    if ($i < scalar @workshop_args) {
+        while ($i < scalar @workshop_args) {
+            workshop_build_image($workshop_args[$i]{'userenv'}, $workshop_args[$i]{'reqs'}, $workshop_args[$i]{'tag'}, $workshop_args[$i]{'skip_update'});
+            push_local_image($workshop_args[$i]{'tag'});
```
k-rister, line 744: Should we be checking for errors here or are we just doing hard exits on errors and calling that good?

rickshaw-run
```diff
@@ -321,38 +324,299 @@ sub dir_entries {
...
+                             . " " . $req_args
+                             . " --proj " . $run{'reg-proj'}
+                             . " --label " . $run{'reg-label'}
+                             . " --tag " . $image_tag;
+    printf "Going to generate a new client-server container image with this workshop cmd:\n\n %s\n", $workshop_build_cmd;
+    printf "This may take a few minutes\n";
+    ($workshop_build_cmd, my $workshop_output, my $workshop_rc) = run_cmd($workshop_build_cmd);
+    my @workshop_output = split(/\n/, $workshop_output);
+    my $workshop_output_file = $run_dir . "/workshop." . $image_tag . ".stdout.txt";
+    my $fh = open_write_text_file($workshop_output_file) || die "Failed to open " . $workshop_output_file . ".out for writing\n";
```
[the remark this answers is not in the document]
atheurer: I believe we automatically add the extension, but I'll check

rickshaw-run
atheurer: I think we just have a die() in these functions we call

k-rister (commented):
I think you should clean up the output some, if possible.  Looking at the CI runs I just see a lot of thought fragments and blank lines that don't make much sense to me.  It seems like a lot of debug output that won't mean anything to someone besides the person who developed this stuff.  And given the questions I answer from our users I think superfluous output should be avoided, if possible.

atheurer:
moving to new PR

PR #231, opened 2021-12-16
multi-server per client support

atheurer:


k-rister (approved):
Shockingly simple.  I guess the postprocessing stuff `just works` since we are aggregating everything anyway, right?

atheurer:
> Shockingly simple. I guess the postprocessing stuff `just works` since we are aggregating everything anyway, right?

Post-processing should work, but in this case (trafficgen) we have no server post-processing anyway.  I was more surprised that the roadblocks just work.

PR #233, opened 2021-12-18
Add quiet mode to ssh commands for endpoints

atheurer:
-This does not affect k8s yet, as we have 'ssh' all over the
 place, and we (a) need to move do_ssh() to base file, and (b)
 have k8s use do_ssh if possible.#

### DIFF
--- a/endpoints/base
+++ b/endpoints/base
@@ -23,6 +23,7 @@ num_clients=0
 num_servers=0
 abort=0
 max_sample_failures=""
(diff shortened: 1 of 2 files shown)


k-rister (approved):
So these flags were only being applied to remotehost endpoint and not k8s endpoint?

atheurer:
> So these flags were only being applied to remotehost endpoint and not k8s endpoint?

Yes.  I'll do a different PR for k8s

PR #240, opened 2022-01-31
add support for disabling tool collection on k8s worker nodes

k-rister:
- use workers-tool-collect:<0|1> for this functionality


atheurer (approved):

rafaelfolco (dismissed):
lgtm

userenvs/stream8.json
```diff
@@ -5,11 +5,11 @@
...
   "userenv": {
-    "name": "stream",
-    "label": "CentOS Stream",
+    "name": "stream8",
+    "label": "CentOS Stream 8",
     "origin": {
-      "image": "registry.centos.org/centos/centos",
-      "tag": "8"
+      "image": "quay.io/centos/centos",
+      "tag": "stream8"
```
atheurer, line 12: Does a stream9 or "latest" exist?

k-rister: This seems "complicated".  There is a stream9 tag and a latest tag...but the latest tag hasn't been updated in quite awhile (1+ year) an even the stream8 tag is looking a bit dated at this point (3 months).  I figured that once we resolved the current problems that we would probably add a stream9 userenv, but I figured I shouldn't complicate the current fixes by doing that right now.

You can see the tags and their age here:

https://quay.io/repository/centos/centos?tab=tags

userenvs/stream8.json
```diff
@@ -5,11 +5,11 @@
     }
   },
   "userenv": {
-    "name": "stream",
-    "label": "CentOS Stream",
+    "name": "stream8",
```
rafaelfolco, line 8: what does happen if we stay on "stream" instead of "stream8" ? Arer you doing this to stay predictable on centos8 replacement ?

userenvs/alma8/userenv.json
```diff
@@ -0,0 +1,14 @@
+  "userenv": {
+    "name": "alma8",
```
rafaelfolco, line 2: just curious if alma is better than rocky for replacing centos

rafaelfolco (approved):
lgtm, are you also planning to remove centos8 requirements from all userenvs ?

userenvs/alma8/userenv.json
k-rister: I did a little research on this and decided it was likely a wash, but for some reason Alma seemed like a more comfortable change to me.

userenvs/stream8.json
k-rister: I figured that we wanted to be more explicit in what we used and given the staleness of the non-versioned tags I figured that using them was probably unwise.

PR #241, opened 2022-02-02
show information (tags) about already built container image stages

k-rister:
- this could come in handy when debugging container image build issues

### DIFF
--- a/rickshaw-run
+++ b/rickshaw-run
@@ -857,6 +857,11 @@ sub source_container_image {
         # image needs to be deleted later, so save a reference to it here
         push(@local_images, $userenv_image);
     }
+    my $x = 0;
+    while ($x < $i) {
+        printf "Processing stage %d (%s)...\n\tReady\n", $x + 1, $workshop_args[$x]{'tag'};
+        $x++;
+    }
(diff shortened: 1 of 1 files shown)


rickshaw-run
```diff
@@ -857,6 +857,11 @@ sub source_container_image {
         # image needs to be deleted later, so save a reference to it here
         push(@local_images, $userenv_image);
     }
+    my $x = 0;
+    while ($x < $i) {
+        printf "Processing stage %d (%s)...\n\tReady\n", $x + 1, $workshop_args[$x]{'tag'};
```
rafaelfolco, line 862: nit: $x + 1 ->  why do you iterate from 0 to $i and print $x + 1 ?

rafaelfolco (approved):

rickshaw-run
k-rister: It is fitting into the existing logic that surrounds this.  I think the rational is that its presenting to the user as layer 1, layer 2, layer 3, etc. while internally the array starts at 0.

https://github.com/perftool-incubator/rickshaw/blob/master/rickshaw-run#L822-L873

PR #243, opened 2022-02-07
add Centos Stream 9 (stream9) userenv

k-rister:


userenvs/stream9.json
```diff
@@ -0,0 +1,94 @@
+{
+  "workshop": {
+    "schema": {
+      "version": "2020.03.02"
+    }
+  },
+  "userenv": {
+    "name": "stream8",
```
HughNhan, line 8: Should it be stream9?

k-rister: It should, editing fail on my part.  Nice catch!

userenvs/stream9/userenv.json
```diff
@@ -0,0 +1,14 @@
+  "userenv": {
+    "name": "stream8",
```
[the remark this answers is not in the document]
k-rister: FYI, this file is actually used to generate the other file you commented on -- so this file is where the actual problem lies.

k-rister:
> ship it!

lol...its kinda funny that you already reviewed it, but I'm glad you did since you got my mistake.  I honestly have no idea if this is correct and I was going to let CI decide if it so it will be very interesting to see what thee results of the CI tests are (this is why I put the PR in draft status).

PR #248, opened 2022-03-01
Move engine back into rickshaw

atheurer:
-Coordinting changes between rickshaw-run, endpoint scripts,
 and client-server-script really needs to be done within a
 single project.


rickshaw-run
```diff
@@ -694,6 +694,7 @@ sub build_reqs {
         push (@$req_ref, "--requirement " . $run{'roadblock-dir'} .  "/workshop.json");
     }
     push (@$req_ref, "--requirement " . $run{'engine-dir'} . "/workshop.json");
```
k-rister, line 696: This should be removed, right?

atheurer: Yes

k-rister (approved):

PR #255, opened 2022-03-22
Keep better track of created pods

atheurer:
-Required to delete them properly


k-rister (approved):
I see that a couple of tests have already passed so you may have actually written a clean patch without testing -- congratulations!

Logic looks sound to me and is basically what I had considered doing to fix it.  I just wasn't sure if I wasn't understanding a reason why you were clearing out the list.

PR #257, opened 2022-03-23
Reorder document indexing

atheurer:
Indexing of ES docs happens only at the very end, so that all information these documents inherit from parent docs is actually available.  If we do not do this, some fields will be empty because the value is not determined until all data for a run has been processed.  An example is begin and end for the run document.

### DIFF
--- a/rickshaw-index
+++ b/rickshaw-index
@@ -264,9 +264,7 @@ sub index_es_doc {
 
 sub index_queued_es_docs {
     foreach my $this_doc (@queued_docs) {
-        print "this_doc:\n";
-        print Dumper $this_doc;
-        index_es_doc(%$this_doc{'doc-type'}, %$this_doc{'iter'}, %$this_doc{'sample'}, %$this_doc{'period'});
+        index_es_doc($$this_doc{'doc-type'}, $$this_doc{'iter'}, $$this_doc{'sample'}, $$this_doc{'period'});
     }
 }
 
@@ -317,7 +315,7 @@ sub write_es_doc {
 
 sub write_queued_es_docs {
     foreach my $this_doc (@queued_docs) {
-        write_es_doc(%$this_doc{'doc-type'}, %$this_doc{'dir'}, %$this_doc{'iter'}, %$this_doc{'sample'}, %$this_doc{'period'});
+        write_es_doc($$this_doc{'doc-type'}, $$this_doc{'dir'}, $$this_doc{'iter'}, $$this_doc{'sample'}, $$this_doc{'period'});
     }
 }
 
--- a/rickshaw-index
+++ b/rickshaw-index
@@ -45,6 +45,7 @@ my $coder = JSON::XS->new->canonical;
 my $result_schema_file;
 my $bench_metric_schema_file;
 my $file_rc;
+my @queued_docs;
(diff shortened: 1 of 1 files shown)


rickshaw-index
```diff
@@ -802,16 +835,18 @@ if (exists $result{'iterations'}) {
                                                     if (! defined $result{'end'} or $result{'end'} < $$this_sample{'periods'}[$period_idx]{'end'}) {
                                                         $result{'end'} = $$this_sample{'periods'}[$period_idx]{'end'};
                                                     }
-                                                    index_es_doc("period", $iter_idx, $sample_idx, $period_idx);
-                                                    write_es_doc("period", $run_dir . "/" . $this_samp_dir, $iter_idx, $sample_idx, $period_idx);
+                                                    queue_es_doc("period", $run_dir . "/" . $this_samp_dir, $iter_idx, $sample_idx, $period_idx);
+                                                    #index_es_doc("period", $iter_idx, $sample_idx, $period_idx);
+                                                    #write_es_doc("period", $run_dir . "/" . $this_samp_dir, $iter_idx, $sample_idx, $period_idx);
```
k-rister, line 840: I would just get rid of this before it becomes cruft...

rickshaw-index
```diff
@@ -802,16 +835,18 @@ if (exists $result{'iterations'}) {
...
                                             }
                                         }
                                     } #cs_ids
                                 } #opendir csnames
                             } #cs_names
-                            index_es_doc("sample", $iter_idx, $sample_idx);
-                            write_es_doc("sample", $run_dir . "/" . $this_samp_dir, $iter_idx, $sample_idx);
+                            queue_es_doc("sample", $run_dir . "/" . $this_samp_dir, $iter_idx, $sample_idx);
+                            #index_es_doc("sample", $iter_idx, $sample_idx);
+                            #write_es_doc("sample", $run_dir . "/" . $this_samp_dir, $iter_idx, $sample_idx);
```
k-rister, line 849: same

rickshaw-index
```diff
@@ -826,8 +861,9 @@ if (exists $result{'iterations'}) {
                 }
                 $$iter{'primary-metric'} = $primary_metric;
                 $$iter{'primary-period'} = $primary_period;
-                index_es_doc("iteration", $iter_idx);
-                write_es_doc("iteration", $run_dir . "/" . $this_iter_dir, $iter_idx);
+                queue_es_doc("iteration", $run_dir . "/" . $this_iter_dir, $iter_idx);
+                #index_es_doc("iteration", $iter_idx);
+                #write_es_doc("iteration", $run_dir . "/" . $this_iter_dir, $iter_idx);
```
k-rister, line 866: same

rickshaw-index
```diff
@@ -836,6 +872,20 @@ if (exists $result{'iterations'}) {
...
+        #index_es_doc("tag", $tag_idx);
+        $tag_idx++;
+    }
+}
+print "Indexing run doc\n";
 index_es_doc("run");
+print "Indexing queued docs\n";
+index_queued_es_docs();
+print "Writing queued docs\n";
+write_queued_es_docs();
```
k-rister, line 888: isn't this backwards?

atheurer: order does not matter here

k-rister (approved):

PR #264, opened 2022-04-28
use namespace for crucible runs

mrbojangles3:
I would like this go through the CI, so I can see if I am missing any cases. After that I would like to discuss if we want to rely on the namespace when we are grepping for pods etc. So we can clean up more of the calls.

### DIFF
--- a/endpoints/k8s/k8s
+++ b/endpoints/k8s/k8s
@@ -575,6 +575,7 @@ function build_pod_spec() {
     echo "  \"kind\": \"Pod\"," >>$json
     echo "  \"metadata\": {" >>$json
     echo "    \"name\": \"$pod_prefix-$name\"" >>$json
+    echo "    \"namespace\": \"$project_name\"" >>$json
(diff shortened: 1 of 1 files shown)


atheurer (approved):
Seems logical to me, just wondering if there's something that breaks when not using the default namespace.

k-rister:
I think we should consider putting a unique identifier (like the SESSION_ID) in the project name -- that would allow multiple Crucible run commands (either from the same controller or different controllers) to be run against the same K8S cluster at the same time without collision occurring.  We are already doing this all over the controller code to allow simultaneous commands to run where we create named pods -- run `git grep SESSION_ID` in the Crucible project to see what I mean.

It also provides linkage to the console log and the run because the SESSION_ID is the run-id and this could come in handy for debugging purposes.

On the flip side, if we uniquely name the project on every run then we won't automatically be cleaning up the resources created from a prior run....but I think we need to figure out how to make all of this work eventually somehow.

atheurer (approved):

PR #272, opened 2022-06-12
skip k8s svc creation when hostNetwork is enabled

HughNhan:


k-rister:
@mrbojangles3 @atheurer Is this ok?

HughNhan:
For got to include the details:
When a pod in hostNework mode, the pod uses the node's networking namespace (stack) directly. There is no need to create a Service for it. As a matter of fact the service will have no endpoint to attach to as the pod's endpoint is the node IP.

Unit Test:
-----
bench-iperf completes all modes: INGRESS, EGRESS, HOST-NETWORK, INTER-HOST, INTER-NODE, INTRA-NODE  successfully.

atheurer (approved):

PR #273, opened 2022-06-15
Dev kmr

k-rister:


endpoints/remotehost/remotehost
```diff
@@ -193,7 +193,7 @@ function process_remotehost_opts() {
                 ;;
             osruntime)
                 case "$val" in
-                    builtin|chroot|podman)
```
rafaelfolco, line 195: and where is the actual code for this option ? not implemented ?

rafaelfolco (approved):
lgtm

endpoints/remotehost/remotehost
k-rister: pretty much, it's the same as some of the other cruft I have been removing...this was used early in the bootstrapping of Crucible/rickshaw and is just totally stale/dead at this point...I couldn't find any code to go along with the option

PR #277, opened 2022-06-22
Add chroot mount per client/server

atheurer:


atheurer:
> I take it this means that it was only appearing to work at lower counts, right?

Yes, I assume because we did not have some shared resources colliding when running only a few client-server-scripts on the same FS.  We never really intended client-server-script to run in a shared FS, but it just got "lucky" some times.  We probably could make it do that, but I just assume give each client-server-script their own FSto run in.

k-rister (approved):

PR #278, opened 2022-06-30
bugfix: only remove $remote_dir on remotehost:chroot after collecting…

k-rister:
… all engine logs


rafaelfolco (approved):
lgtm I just don't get why only chroot case needs removal

k-rister:
> lgtm I just don't get why only chroot case needs removal

podman case is self contained and the container removal cleans up everything

chroot is a much more manually configured and pieced together entity that we have to properly clean up all the various mounts and files that we create ourselves

PR #282, opened 2022-07-19
properly pass the abort signal back from the engines

k-rister:
- previous updates broke the path which passed the signal back so
  reconnect things


atheurer (dismissed):

k-rister:
I think there is a little bit more todo here, looking into it today.

atheurer (approved):

PR #284, opened 2022-07-21
change 'client-server' and 'client_server' to 'engine'

k-rister:
- the engine code is so much more than just clients or servers at this
  point that this labeling is misleading

- it is also confusing to be talking about clients and/or servers and
  then talking about client-server...they can be and are significantly
  different things, so separate the terminology to make discussion and
  documentation clearer


endpoints/k8s/k8s
```diff
@@ -116,7 +116,7 @@ function endpoint_k8s_test_start() {
     # the client is in relation to the server.  If the client is within the same k8s cluster,
     # we create a k8s-service, so the client can use an IP which is more persistent
     # than a pod's IP (this allows pods to come and go while keeping the same IP).  This is not
-    # absolutely necessary for client-server benchnarks, but it is a best practice for cloud-native
+    # absolutely necessary for our  benchmarks, but it is a best practice for cloud-native
```
HughNhan, line 119: Cool. Not all are :1,$ /client-server/engine/g'

engine/engine-script
```diff
@@ -7,21 +7,21 @@ echo "BASH_VERSION=${BASH_VERSION}"
 
 test -e /etc/profile && . /etc/profile
 
-if [ ! -e /usr/local/bin/client-server-script-library ]; then
-    echo "ERROR: Could not location client-server-script-library"
+if [ ! -e /usr/local/bin/engine-script-library ]; then
+    echo "ERROR: Could not location engine-library"
```
HughNhan, line 11: engine-script-library?

k-rister: yeah, odd naming...I failed to come up with anything better I suppose:

k-rister: https://github.com/k-rister/rickshaw/tree/dev-kmr/engine

atheurer (approved):

PR #295, opened 2022-08-09
Switch almalinux image to quay.io

rafaelfolco:


k-rister (approved):

rafaelfolco:
@k-rister Can I merge as is or you prefer the merge commit to be squashed ?

k-rister:
I'm fine with as is.

PR #315, opened 2022-09-22
Support adding benchmark name to bench-metric

atheurer:
-this schema is used for post-process-data.json
-when rickshaw supports just 1 benchmark type per run,
 the benchmark name was generally already known,
 as post-processing was done for a known, singular
 benchmark.  To prepare to multi-benchmark use, we
 need the post-process-data to include the benchmark
 name.  Once this is merged, all benchmark subprojects
 will need to be updated to include their benchmark
 name in this file they generate.

### DIFF
--- a/schema/bench-metric.json
+++ b/schema/bench-metric.json
@@ -23,6 +23,9 @@
       ],
       "additionalProperties": false
     },
+    "benchmark": {
+      "type": "string"
+    },
     "primary-metric": {
       "type": "string"
     },


k-rister (approved):
I'm almost positive that while this will allow the new parameter to exist it will not require it.  I recently noticed this when doing some other work and it is on my todo list to fix.  Since it's such a rampant problem at this point in the rickshaw schemas there is no reason to worry about it for this one thing, besides, if we made this actually be required it would make CI fail since the benchmarks wouldn't have it yet, and for that matter it would be the usual chicken-and-egg where you could not add it to the benchmarks without it being added here first either.

atheurer:
> I'm almost positive that while this will allow the new parameter to exist it will not require it. I recently noticed this when doing some other work and it is on my todo list to fix. Since it's such a rampant problem at this point in the rickshaw schemas there is no reason to worry about it for this one thing, besides, if we made this actually be required it would make CI fail since the benchmarks wouldn't have it yet, and for that matter it would be the usual chicken-and-egg where you could not add it to the benchmarks without it being added here first either.

This is why I did not make it a requirement yet, because it would break every benchmark.  Once all existing benchmarks have it, then and only then can we change to being required.

PR #316, opened 2022-09-30
Osp endpoint

atheurer:


pradiptapks:
We can use the OpenStack heat service to reduce the complexity of the endpoint to manage the VM resource.
https://docs.google.com/document/d/1wzkgf8Ry1QIW05sAknSKVarV9cC2QpRUn84uck_89jE/edit?usp=sharing

endpoints/osp/osp
```diff
@@ -0,0 +1,582 @@
...
+else
+    echo "Could not find endpoint source file "$endpoint_base_dir/base", exiting"
+    exit 1
+fi
+
+echo "#running osp endpoint"
+clear_all_servers="1"
+chroot_rbind_mounts="proc dev sys lib/modules usr/src boot var/run"
+image_cache_size=3
+default_flavor="nfv-large"
```
k-rister, line 30: this seems like an odd default

endpoints/osp/osp
```diff
@@ -0,0 +1,582 @@
...
+default_availability_zone="nova"
+osruntime="podman"
```
k-rister, line 32: should the remotehost and osp endpoints have the same default osruntime?  right now remotehost is chroot and osp is podman...

endpoints/osp/osp
```diff
@@ -0,0 +1,582 @@
...
+base_osruntime_bootstrap="$endpoint_base_dir/osruntime-bootstrap"
+project_name="crucible-rickshaw"
+new_osp_followers="" # engines which only run tools for openstack nodes
+vm_image="centos9"
+vm_user="centos"
+declare -A cpuPartitioning
+declare -A flavor
+declare -A availability_zone
+declare -A ips
+declare -A active_compute_nodes
```
k-rister, line 45: Since osruntime is not declared here does that mean it can't be changed per engine?

endpoints/osruntime-bootstrap
```diff
@@ -0,0 +1,245 @@
...
+        fi
+    done
+    log_output "validate_vars() end"
+}
+
+validate_vars
+
+image_cache_size=3
+chroot_rbind_mounts="proc dev sys lib/modules usr/src boot var/run"
+host_mounts="run/openvswitch"
```
k-rister, line 92: this seems odd to me...what is OVS isn't present?

k-rister (approved):

endpoints/osp/osp
k-rister: Sorry, I meant the `default_flavor`

endpoints/osruntime-bootstrap
atheurer: The osruntime-bootstrap checks for existence of the directory in the host first, and will not create the mount if it does not exist.

endpoints/osp/osp
k-rister: Is it just a problem with sourcing the env file?

atheurer: Yes, I just have not gotten back to see if sourcing it inside the container works

endpoints/osp/osp
atheurer: Ah, yes, I should probably change that.  However, I am not sure what value to use.  Perhaps we set it to "default", which IMO is totally useless, but then when it checks for the flavor and errors out, the user will then know they need to provide the flavor.

endpoints/osp/osp
atheurer: yes

endpoints/osp/osp
k-rister: No idea if this is representative or not, but in my limited work with `microstack` so far I'm finding `m1.small` to be the minimum required flavor (`m1.tiny` does not create a disk big enough for the the default Centos Stream 9 and Ubuntu 22.04 images I downloaded).

endpoints/osp/osp
atheurer: Sourcing this file would not work, so reverted back to adding cmdline args to boostrap

k-rister (approved):

PR #340, opened 2022-11-16
Limit fork bombs

atheurer:


k-rister (approved):

k-rister:
I believe this PR is no longer applicable and it is in conflict with the current upstream code.

PR #343, opened 2022-12-02
add stream8-flexran env

HughNhan:
Beginning from FlexRAN v22.03 going forward, we build the flexran client-server image on top of a stream8-flexran based image. The instruction to build the based image is [here](https://docs.google.com/document/d/1sKJzcm0oiui5C6e6QzHLFMAX8HAzAnUmQwD7-kAcpcM/edit#heading=h.bs1eewbj7n2w).

Prior to this, the engines have to mount the host filesystem that contains the BIG flexran blob (100MB).

### DIFF
--- a/userenvs/stream8-flexran.json
+++ b/userenvs/stream8-flexran.json
@@ -1,0 +1,94 @@
+{
+  "workshop": {
+    "schema": {
+      "version": "2022.07.25"
+    }
+  },
+  "userenv": {
+    "name": "stream8-flexran",
+    "label": "flexRAN Stream 8",
+    "origin": {
+      "image": "quay.io/xxx/yyy",
+      "tag": "zzz"
+    },
+    "properties": {
+      "packages": {
+        "type": "rpm",
+        "manager": "dnf"
+      }
+    }
+  },
+  "requirements": [
+    {
+      "name": "python39",
+      "type": "distro",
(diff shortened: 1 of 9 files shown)


userenvs/stream8-flexran.json
```diff
@@ -0,0 +1,94 @@
...
+    "schema": {
+      "version": "2022.07.25"
+    }
+  },
+  "userenv": {
+    "name": "stream8-flexran",
+    "label": "flexRAN Stream 8",
+    "origin": {
+      "image": "quay.io/xxx/yyy",
+      "tag": "zzz"
```
k-rister, line 12: Is the implication here that the user is supposed to update these fields prior to using this?

HughNhan: Yes.  Intel seems to crank out 2,3 FlexRAN releases a year.  We build images from these releases as needed. So we are on a moving train.

k-rister: My concern, if you want to call it that, is not so much that the "train" is moving (you could say that about all of the userenvs really) but that it needs to be well documented somewhere/somehow that there is this expectation for this one userenv.  Maybe add a `README.md` file in the `stream8-flexran` directory that documents how this userenv is expected to be used?

k-rister (approved):

PR #351, opened 2022-12-09
switch to reusable crucible-ci core workflow

k-rister:


rafaelfolco (changes requested):
could you please break down into 2 commits? the one unrelated to the reusable ci might be good to be a separate one, in case you need to revert.

k-rister:
> could you please break down into 2 commits? the one unrelated to the reusable ci might be good to be a separate one, in case you need to revert.

Thank you for catching this, I accidentally included a bunch of stuff that I did not intend to.  I will fix it up.

rafaelfolco (approved):

Issue #354, opened 2023-01-11
include lib/firmware in bind mounts


atheurer:
Needed for some programs, like DPDK, to load firmware.

k-rister:
Already done.

PR #355, opened 2023-01-11
add support for automatically bind mounting /lib/firmware from the ho…

k-rister:
…st into the engine(s)


k-rister:
Addresses #354

bmichalo:
I tried a crucible dpdk-testpmd application run with an E810 adapter and it worked.  Before this patch, it failed.

atheurer (approved):

PR #356, opened 2023-01-12
fix prescript for single-bench run

HughNhan:
Not sure if this is a good fix, but it makes single bench to be able to use the prescript mechanism.
I have not tried multi-bench prescript.  Hope you can test or suggest how to test it.


k-rister:
I'm fine with this and require it for certain workloads myself.  That said, there really needs to be a more complete solution to this problem designed (it has limitations as is).  I'm also not aware if there are any downsides to doing this.  Thoughts @atheurer ?

k-rister:
Closing this PR due to branch conflicts.  If this is still needed (and I think it is) we need to file a new PR.

PR #362, opened 2023-02-06
Add support for unbounded workloads

k-rister:


engine/engine-script-library
```diff
@@ -943,20 +949,28 @@ function process_bench_roadblocks() {
...
                 fi
 
-                run_bench_cmd "client" "client" "${abort}" "${abort}" "0" "${start_cmd}"
-                abort_rc=$?
-                if [ "${abort_rc}" == "0" ]; then
-                    abort_arg=""
+                if [ "${unbounded_roadblock}" == "yes" ]; then
+                    rb_name=$(printf "%s%s" ${rb_prefix} "client-start-end")
+                    prepare_roadblock_user_msgs_file "${iter_samp_dir}" "${rb_name}"
+                    do_roadblock "${rb_name}" ${timeout} wait-for "/usr/local/bin/engine-script-library run_bench_cmd '${cs_type}' 'client' 'client' '${abort}' '${abort}' '0' '${start_cmd}'"
```
rafaelfolco, line 955: timeout still exists here, so what is the effect of unbounded ?

rafaelfolco (approved):
lgtm, w/ question

engine/engine-script-library
k-rister: The timeout is in effect until the initial connections are made.  Once that happens then instead of waiting for the timeout to occur it waits for the forked wait-for process to complete, this is currently not bounded by any time constraints and will allow the forked process to determine how long all the participants wait, hence the term "unbounded".  NOTE:  While waiting there is a background heartbeat being sent to all participants from the controller and if one of the participants fails to reply to the heartbeat then the roadblock will fail -- maintaining the the ability to ensure that all participants are still properly responding and have not become "lost".

rafaelfolco (approved):

PR #366, opened 2023-02-13
Dev kmr

k-rister:


k-rister:
> don't you need to remove ci jobs first ?

The ci jobs are automatically generated from the source: https://github.com/perftool-incubator/crucible-ci/tree/main/.github/actions/get-userenvs

PR #373, opened 2023-02-27
change controller ping behavior

k-rister:
- use ping deadline

- do not treat a ping failure as an error state

  - the ping is simply for informational purposes in debugging situations

  - roadblock should have sufficient retry and logging to properly handle network connectivity problems -- if that is not true it is a roadblock bug that should be fixed there, not by the caller

### DIFF
--- a/endpoints/base
+++ b/endpoints/base
@@ -568,7 +568,7 @@ function do_roadblock() {
     local msgs_log_file="$roadblock_msgs_dir/$label.json"
     local cmd=""
     local role="follower"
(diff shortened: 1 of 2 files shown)


rafaelfolco (approved):
lgtm

mrbojangles3:
I agree that roadblock is the right place for retry logic. Do we want to try to wrap the ping call in some kind of debug flag? and only do it if we are working to solve a crucible problem? These pings did help me diagnose a performance problem so I am fond of  them.

PR #380, opened 2023-04-05
Alias target config 'all' to 'default'

rafaelfolco:
'all' is an alias for default, which tells rickshaw to apply a config to all of the targets.


rafaelfolco:
hopefully this fixes problem at https://github.com/perftool-incubator/crucible-ci/actions/runs/4619083415/jobs/8167390493?pr=97
from PR https://github.com/perftool-incubator/crucible-ci/pull/97

k-rister (approved):

PR #382, opened 2023-04-14
copy both roadblocker.py and roadblock.py when necessary

k-rister:


k-rister:
This is a prereq for https://github.com/perftool-incubator/roadblock/pull/53

rafaelfolco (approved):
lgtm

PR #383, opened 2023-04-17
Create engine-script-library.py

grdumas:
I created an outline for the engine-script-library script conversion and attempted to implement a few functions. I'm unsure about my version of the do_roadblock function: any suggestions for how I might test and verify?

PR #384, opened 2023-04-17
temporarily bypass the "real CI" for the engine python port work

k-rister:
- use the faux CI for now


k-rister:
@grdumas we need to merge this so that your work (such as #383 ) doesn't cause needless CI runs.  Please review this ASAP.

PR #388, opened 2023-04-24
Write JSON blk setting as str

rafaelfolco:
json.dumps adds unwanted chars


rafaelfolco:
https://github.com/perftool-incubator/rickshaw/pull/386 wasn't enough since json.dumps messes up with the json str by adding escape chars and replaces double quotes with single ones.

k-rister (approved):

PR #392, opened 2023-05-08
Revision of do_roadblock()

grdumas:


engine/engine-script-library.py
```diff
@@ -33,104 +33,124 @@ def abort_error(msg, sync):
...
+    msgs_log_file = os.path.join(roadblock_msgs_dir, f"{label}.json")
+
+    # Initialize the command string and set the role to "follower"
+    cmd = ""
     role = "follower"
 
-    # Replace rickshaw_host with the appropriate hostname or IP address
-    rickshaw_host = "localhost"
+    # Use subprocess to run a "ping" command to the rickshaw host
     subprocess.run(["ping", "-w", "10", "-c", "4", rickshaw_host])
```
k-rister, line 79: I've been looking into [Fabric](https://www.fabfile.org/)/[Invoke](https://www.pyinvoke.org/) for some similar stuff, what do you think of them instead of using subprocess directly?  I'm mainly curious here with this question...

k-rister (approved):

PR #393, opened 2023-06-07
Multibench singlejson

rafaelfolco:


rafaelfolco:
Still need to make the mv-params config return a single object from the array, or itself if an object type. Default index is 0 in case of an array type.

rafaelfolco:
@k-rister & @atheurer this is the PR that integrates the single-json (all in one json) with multibench. Tests and CI job to be created in a separated patch to simplify the complexity of changing schemas and not breaking the current functionality.

PR #399, opened 2023-08-30
Add multibench support to all-in-one json

rafaelfolco:
Add benchmarks block to identify bench ids

PR #400, opened 2023-09-11
Add multibench support to all-in-one json

rafaelfolco:
Add benchmarks block to identify bench ids.
The blockbreaker utility extracts the benchmark
from the runfile and processes multibench config.

This requires additional changes on the crucible
code to fully support becnhamrks from the json
runfile.


rafaelfolco:
attempt to solve permission issues seen at #399

PR #406, opened 2023-09-27
Dev kmr

k-rister:


util/blockbreaker.py
```diff
@@ -24,7 +24,7 @@ def process_options():
                         dest = "config",
                         required = True,
                         help = "Configuration type to get from the json file",
-                        choices = [ "benchmarks", "mv-params", "tool-params", "passthru-args", "tags", "endpoint", "endpoints" ])
+                        choices = [ "benchmarks", "mv-params", "tool-params", "run-parameters", "tags", "endpoint", "endpoints" ])
```
rafaelfolco, line 27: nit... in case you need a respin, I'd prefer "runtime-params" to be more like the others

util/blockbreaker.py
```diff
@@ -177,18 +177,22 @@ def json_to_stream(json_obj, cfg, idx):
...
+                            stream += key + ':'
+                        stream += val_str + ','
                 except:
                     raise Exception("Error: Unexpected object type %s" % (type(val)))
                     return None
 
-    # remove last ","
-    if len(stream)>0:
-        stream = stream[:-1]
+    if cfg != 'run-parameters':
```
rafaelfolco, line 192: this is fine, you could also replace "," with ""

rafaelfolco (approved):
lgtm, comments are just non-required

util/blockbreaker.py
k-rister: Hmm, I could see using `run-params` as that would align with the other values and now that you mention it that kinda seems like a good idea.  IMHO, 'runtime' kinda implies something about the actual benchmark run execution time which is not really what these parameters are about.

rafaelfolco (approved):

PR #410, opened 2023-10-04
Add utility to validate the run-file

rafaelfolco:
And update doc

### DIFF
--- a/util/README.md
+++ b/util/README.md
@@ -1,29 +1,42 @@
 
-### JSON Run File
+# Crucible run-file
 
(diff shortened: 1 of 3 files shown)


k-rister (dismissed):
My only concern is if `validate.py` is a bit generic name-wise, should the name be a bit more descriptive of what it's validating?

rafaelfolco:
> My only concern is if `validate.py` is a bit generic name-wise, should the name be a bit more descriptive of what it's validating?

ack

PR #426, opened 2023-11-03
Support Common Data Model "v7dev"

atheurer:
A number of new "names" in metric-desc ES doc
Implement passing info to engines via engine-init roadblock
Determine that info in remotehost and k8s endpoints, to pass to engines
Osp not implemented yet

### DIFF
--- a/rickshaw-index
+++ b/rickshaw-index
@@ -352,8 +352,9 @@ sub wait_for_metric_descs {
 # with create_es_doc().  Metrics can be indexed from either a benchmark sample directory or a tool
 # directory.
 sub index_metrics {
-    my $index_or_queue = shift;
-    my $metr_file = shift; # filename without .json or .csv
+    my $index_or_queue = shift; # what action to take, index = submit to ES, queue = enqueue to file for bulk index later
+    my $metr_dir = shift; # directory where metric files exist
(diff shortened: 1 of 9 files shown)


rickshaw-index
```diff
@@ -364,6 +365,60 @@ sub index_metrics {
...
+            my @varnames = ('HOSTNAME', 'engine_type', 'engine_role', 'benchmark_group', 'benchmark_role', 'hosted_by', 'hypervisor_host', 'osruntime', 'endpoint_label', 'userenv');
+            while (<$eng_env_fh>) {
+                chomp;
+                foreach my $varname (@varnames) {
+                    if (/^$varname=(.*)$/) {
+                        my $val = $1;
+                        $varname =~ s/_/-/g;
+                        if ($varname eq "HOSTNAME") {
+                            $varname = "hostname";
+                        }
```
k-rister, line 412: Why the special handling for this one?

k-rister (dismissed):

rickshaw-index
atheurer: only because HOSTNAME happens to be defined in upper case (we don't define it) from env vars, and all the metric-desc 'names' are usually lower case.

k-rister (approved):

PR #429, opened 2023-11-27
One tool

atheurer:
This PR will have multiple commits for different areas of work:

1) Enabling remotehost endpoint to launch one tool per engine,
    plus any changes needed in rickshaw-*:
    - Add communication to endpoiont-deploy endpoint to receive messages about new roadblock followers from more than 1 endpint
2) Enabling k8s endpoint to launch one tool per engine
3) Enabling openstack endpoint to lunch one tool per engine
4) Having rickshaw-run source/build an image per tool for the run
5) Having all endpoints use the specific image for each tool they launch


endpoints/remotehost/remotehost
```diff
@@ -304,30 +381,40 @@ function launch_osruntime() {
 
     set_total_cpupart
 
+    # Launch the tool engines (support only pods)
+    local tools=`awk -F: '{print $1}' $config_dir/tool-cmds/profiler/start`
+    echo "Creating a pod for each of these tools: $tools"
+    tool_count=1
+    echo "Going to create a pod for each tool"
+    for tool in $tools; do
+        # This label will not work with mulitple remotehost endpoints!
```
k-rister, line 390: What do we have to do to fix this comment?  I'm actively using multiple remotehost endpoints right now so I guess this would break me?

rickshaw-run
```diff
@@ -34,7 +34,7 @@ use toolbox::logging;
 use toolbox::run;
 use toolbox::jsonsettings;
 
-$toolbox::logging::debug = 0;
+$toolbox::logging::debug = 1;
```
k-rister, line 37: We don't want to merge this change.  I assume this was for debugging purposes...

rickshaw-run
```diff
@@ -896,7 +896,7 @@ sub source_container_image {
         if ($count == 0) {
             $userenv_arg = " --userenv " . $rickshaw_project_dir . "/userenvs/" . $userenv . ".json";
             $req_arg = "";
-            $skip_update = "false";
+            $skip_update = "true";
```
k-rister, line 899: Not sure we want to merge this either...

rickshaw-run
```diff
@@ -2168,12 +2168,12 @@ sub process_roadblocks() {
     if (defined $messages_ref) {
         foreach my $msg (@{ $$messages_ref{'received'} })  {
             if (exists $$msg{'payload'}{'message'}{'user-object'}{'new-followers'}) {
-                @new_followers = @{ $$msg{'payload'}{'message'}{'user-object'}{'new-followers'} };
-                printf "Endpoint created followers: %s\n", join(" ", sort(@new_followers));
-                last;
```
k-rister, line 2173: I suppose you meant to drop this `last` call?

k-rister (changes requested):
So does this completely eliminate the concept of the user creating a profiler engine themselves?  I see a couple places where you removed it from a set of roles that are being checked for so I'm guessing it does.  If so, we are probably going to have CI issues with this that I will need to resolve since we explicitly test profiler engine creation.

Also, I think I identified a couple of local changes that slipped in that we don't actually want to merge upstream.

endpoints/remotehost/remotehost
atheurer: This is odd.  I thought I included a newer commit that fixes this. FWIW this is all draft

atheurer: And yes I do intend to support multiple remotehost endpoints to same host

atheurer: I think I just forgot to remove the comment which no longer applies

rickshaw-run
atheurer: Yes, absolutely.  I am not even sure why it was there in the first place, but now more than ever we have to process extra followers from all endpoints.

endpoints/remotehost/remotehost
atheurer: BTW it looks like multiple osp or k8s endpoints are also broken, not from this code but just since they have existed.  For example, in the case of mulitple k8s, we would end up having a worker-1 engine from both endpoints.

rickshaw-run
atheurer: yes

k-rister:
> I must be inadvertently removing something I did not understand had a purpose. Is there a specific example of a crucible run that has a user creating a profiler engine? I can then use that to make sure it works here.

https://github.com/perftool-incubator/crucible-ci/blob/main/.github/actions/integration-tests/run-ci-stage1#L388

I think the intended purpose of this functionality was to be able to launch profilers on something like a KVM host, storage server, etc. where you wanted to collect data but weren't actually running a workload.

atheurer:
> > I must be inadvertently removing something I did not understand had a purpose. Is there a specific example of a crucible run that has a user creating a profiler engine? I can then use that to make sure it works here.
> 
> https://github.com/perftool-incubator/crucible-ci/blob/main/.github/actions/integration-tests/run-ci-stage1#L388
> 
> I think the intended purpose of this functionality was to be able to launch profilers on something like a KVM host, storage server, etc. where you wanted to collect data but weren't actually running a workload.

Ahh, OK, I misunderstood your question.  I thought you were referring to some kind of user-provided profiler, which we of course don't have [yet], and so it did not make sense to me.  I'll make sure we can still include remotehost endpoints which are only profilers.  I'll test that next.

engine/engine-script-library
```diff
@@ -1093,6 +1106,10 @@ function stop_tools() {
             while read -u 9 line; do
                 tool_name=`echo $line | awk -F: '{print $1}'`
                 tool_cmd=`echo $line | sed -e s/^$tool_name://`
+                if [ ! -z "$one_tool" -a "$one_tool" != "$tool_name" ]; then
+                    echo "Not starting $tool_name because this engine only runs one tool, $one_tool"
```
k-rister, line 1110: Should be `Not stopping...`

rickshaw-post-process-bench
```diff
@@ -330,5 +325,10 @@ for (my $i = 1; $i <= scalar @{ $run{'iterations'} }; $i++) {
             }
             print "Post-processing complete\n";
         }
+        if (scalar @samp_dirs < $run{'num-samples'}) {
+            printf "Iteration %d, [%s] is considered failed because there were not enough passing samples (%d)\n",
+                   $i, $this_iter_dir, $run{'num-samples'};
+            system("/bin/mv " . $run_dir . "/" . $this_iter_dir . " " . $run_dir . "/" . $this_iter_dir . "-fail");
+        }
```
k-rister, line 332: Is this change to ensure that the post-processing runs before the directory is rename?  Something else?

rickshaw-run
```diff
@@ -96,6 +96,7 @@ my $workshop_roadblock_opt = "";
 my %utility_configs;
 my %tools_configs;
 my @tools_params;
+my $default_tool_userenv = 'fedora39';
```
k-rister, line 99: I wonder if this should be in the settings JSON file.

rickshaw-run
```diff
@@ -979,10 +991,6 @@ sub source_container_image {
         push(@local_images, $userenv_image);
     }
     my $x = 0;
-    while ($x < $i) {
-        printf "Processing stage %d (%s)...\n\tReady\n", $x + 1, $workshop_args[$x]{'tag'};
-        $x++;
-    }
```
k-rister, line 985: I'm pretty sure I added this for a reason, not sure we should get rid of it.

k-rister (commented):
First pass review.  I plan on running some tests with it myself just to better understand it, but I'll get to that later.

endpoints/base
```diff
@@ -305,42 +319,42 @@ function process_common_endpoint_opts() {
...
-                let num_clients=$num_clients+1
-            elif [ "$arg" == "servers" -o "$arg" == "server" ]; then
-                servers[$ids]="server-$ids"
-                let num_servers=$num_servers+1
-            elif [ "$arg" == "profiler" ]; then
-                collectors[$ids]="profiler-$ids"
-                let num_collectors=$num_collectors+1
-            fi
+            echo "This client/server/profiler label is not recognized:"
+            echo "$arg-$val"
```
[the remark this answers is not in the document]
atheurer: actually it probably should be fatal

rickshaw-post-process-bench
atheurer: yes

rickshaw-run
atheurer: I think I was just looking for cleaner output, but I'll put it back.

rickshaw-run
atheurer: moving it

endpoints/k8s/k8s
```diff
@@ -887,18 +778,174 @@ function build_pod_spec() {
...
+            echo '            "name": "hostfs-run"' >>$json
+            echo '          },' >>$json
+            echo '          { "mountPath": "/lib/firmware",' >>$json
+            echo '            "name": "hostfs-firmware"' >>$json
+            echo '          },' >>$json
+            echo '          { "mountPath": "/lib/modules",' >>$json
+            echo '            "name": "hostfs-modules"' >>$json
+            echo '          }' >>$json
+            echo "        ]" >>$json
+        fi
```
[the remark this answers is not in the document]
atheurer: The type is used here to distinguish between a pod that is exclusively for tools vs one that also handles a benchmark, but all engines that run a tool are now engine-type "profiler", regardless of endpoint.  However, since there is a pod name and a container name for OCP, the pod name retains a label like "worker-1", while the container names have "profiler-<endpoint-label>-<tool>-id".

endpoints/k8s/k8s
```diff
@@ -1165,7 +1214,7 @@ function endpoint_k8s_cleanup() {
         echo "Error getting pod logs"
     fi
 
-    return $(( ${log_c} + ${delete_pods_rc} + ${delete_namespace_rc} ))
+    return $(( ${log_rc} + ${delete_pods_rc} + ${delete_namespace_rc} ))
```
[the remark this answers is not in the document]
atheurer: where does log_c come from?

endpoints/osp/osp
```diff
@@ -264,167 +312,224 @@ function create_servers() {
...
-        server_create_cmd="$osp_pre_cmd openstack server create --flavor $this_flavor --nic port-id=$mgmt_port_name --image $vm_image --availability-zone $this_availability_zone --key-name $vm_ssh_key_name --user-data $this_osruntime_bootstrap --key-name $vm_ssh_key_name $vm_name"
-        echo "Going to run via ssh: $server_create_cmd"
-        do_ssh ${user}@${host} $server_create_cmd >${endpoint_run_dir}/openstack-create-$vm_name.txt 2>&1
-        rc=$?
-        if [ $rc -ne 0 ]; then
-            echo "openstack server create exit code: $rc"
-            cat ${endpoint_run_dir}/openstack-create-$vm_name.txt
-            abort_error "`cat ${endpoint_run_dir}/openstack-create-$vm_name.txt`" endpoint-deploy-begin
-        fi
+function create_server() {
```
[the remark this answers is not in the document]
atheurer: Not in the context of openstack.  The resource being created is an openstack server

endpoints/remotehost/remotehost
```diff
@@ -116,14 +142,14 @@ function endpoint_remotehost_test_start() {
                 let count=$count+1
             done
 
-            local server_outside_endpoint=1
+            server_outside_endpoint=1
```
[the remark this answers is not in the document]
atheurer: Style.  As remotehost was almost completely re-written, I decided to document the local and globals at the top of each function.

endpoints/remotehost/remotehost
```diff
@@ -160,62 +210,50 @@ function cleanup_osruntime() {
...
-        if [ "${os_runtime}" == "chroot" ]; then
-            data_file=${endpoint_run_dir}/chroot-container-mount.txt
-            if [ -e ${data_file} ]; then
-                while read tmp_cs_label container_mount; do
-                    if [ "${this_cs_label}" == "${tmp_cs_label}" ]; then
-                        for fs in ${chroot_rbind_mounts}; do
-                            echo "Removing ${fs} from ${container_mount}"
-                            do_ssh ${user}@${host} umount --verbose --recursive ${container_mount}/${fs}
+    for engine_label in ${clients[@]} ${servers[@]} ${profilers[@]}; do
+        set_osruntime_numanode_cpupart $engine_label
```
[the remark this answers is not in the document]
atheurer: Because you need to know the osruntime for each engine you process.  set_osruntime_numanode_cpupart() sets those 3 vars.

endpoints/remotehost/remotehost
```diff
@@ -146,8 +172,32 @@ function endpoint_remotehost_test_start() {
     fi
 }
 
+function manage_image_cache() {
```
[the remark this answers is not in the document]
atheurer: I just went by the comment below it, "Performing container image cache management".  Is there something else you would prefer?

endpoints/remotehost/remotehost
```diff
@@ -282,280 +320,366 @@ function remotehost_req_check() {
...
+    (( packrat_rc += $? ))
 
-    do_ssh ${user}@${host} rm -Rf ${remote_dir}/packrat ${remote_dir}/sysinfo
-    (( packrat_rc += ${?} ))
+    do_ssh $user@$host rm -Rf $remote_dir/packrat $remote_dir/sysinfo
+    (( packrat_rc += $? ))
 
-    return ${packrat_rc}
+    return $packrat_rc
+}
```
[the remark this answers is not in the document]
atheurer: Yes, I am.  IMO the gratuitous use of {}'s assumes a lack of comprehension of the bash syntax by the reader.

endpoints/k8s/k8s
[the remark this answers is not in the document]
k-rister: Sorry, I meant I made an oops, thanks for fixing it.

endpoints/osp/osp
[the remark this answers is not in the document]
k-rister: Hmm, I guess I misunderstood what was happening here.  I thought this was creating engines, my bad.

endpoints/remotehost/remotehost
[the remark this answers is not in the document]
k-rister: Ok, I was just curious because I recall seeing a problem somewhere once when `local` was used on the same line where the variable was assigned a value....I was curious if you were seeing that or not because I never really understood what was happening -- I just remember that separating the use of `local` and the assignment seemed to fix a problem.

endpoints/remotehost/remotehost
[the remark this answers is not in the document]
k-rister: ack

endpoints/remotehost/remotehost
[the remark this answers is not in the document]
k-rister: I think that comment only applied to a single command.  The other commands that came after were just capturing the podman state for debugging purposes and were unrelated to the cache mgmt.

endpoints/remotehost/remotehost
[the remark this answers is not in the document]
k-rister: I'm not complaining, just merely commenting.  I think we have discussed this somewhat before, but IMHO there are places where the use of the `{}` really help in understanding what is going on, especially when you reviewing code that you the reader didn't write so the variable names may not be familiar -- ie. does `$foo_bar` mean `${foo}_bar` or does it mean `${foo_bar}`.  Then, in order to avoid worrying about whether I should be using the `{}` or not in individual situations I just put myself in the habit of always using them so that it's always explicit.

rickshaw-settings.json
```diff
@@ -9,7 +9,7 @@
 	}
     },
     "userenvs": {
-	"default": "rhubi8"
+	"default": "fedora39"
```
k-rister, line 12: Are we sure we want to tie the default userenv for benchmarks to the default userenv for tools?

PR #432, opened 2023-11-28
Use variable, Instead of hardcoded network name during server creation

k-rister:


k-rister:
@rajeshP524   FYI, new PR for your submission.

rafaelfolco (approved):
lgtm

PR #436, opened 2023-12-08
change a debug logging statement to always be shown

k-rister:
- for "large" runs the "organize_run_data" function can take a very long time to execute, make sure the user knows what is going on by always printing this message

### DIFF
--- a/endpoints/base
+++ b/endpoints/base
@@ -25,7 +25,8 @@ user=""
 engine_script_start_timeout=""
 max_sample_failures=""
 total_cpu_partitions=0
-export ssh_opts="-q -o StrictHostKeyChecking=no -o PasswordAuthentication=no"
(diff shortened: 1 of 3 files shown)


endpoints/base
```diff
@@ -25,7 +25,8 @@ user=""
 engine_script_start_timeout=""
 max_sample_failures=""
 total_cpu_partitions=0
-export ssh_opts="-q -o StrictHostKeyChecking=no -o PasswordAuthentication=no"
+export ssh_opts="-q -o StrictHostKeyChecking=no -o PasswordAuthentication=no -o ControlMaster=auto -o ControlPath=~/.ssh/sockets/%r@%h-%p -o ControlPersist=60"
```
rafaelfolco, line 28: I have no idea what this does :)
 %r@%h-%p

rafaelfolco (approved):

endpoints/base
k-rister: IIRC thiis creates or uses a socket with the filename being of the form `<username>@<hostname>-<port>` so that repeated connections to a host with the same username on the same port can reuse the existing connection that created the socket.

PR #437, opened 2023-12-12
update blockbreaker.py and the associated schemas to have consistent/…

k-rister:
…universal handling of numbers/number ranges/list of number ranges

- for example, these would all be treated similarly no matter where they are being specified:

  1
  1-3
  1-3+5-7


rafaelfolco (approved):
lgtm
unrelated catch that comes from the original design: output-* files that are streams and not JSON should not have .json extension (although they are inside JSON/ dir)

k-rister:
> lgtm
> 
> unrelated catch that comes from the original design: output-* files that are streams and not JSON should not have .json extension (although they are inside JSON/ dir)

Good point, I agree, I will fix.

rafaelfolco (approved):

PR #463, opened 2024-02-14
fixes for resources and securityContext

atheurer:


endpoints/k8s/k8s
```diff
@@ -733,14 +733,15 @@ function build_pod_spec() {
...
+        # If user provides pod securityContext
+        if [ "$type" == "cs" ]; then
+            set +u
+            if [ ! -z "${PsecurityContext[$name]}" ]; then
+                echo -e ",${PsecurityContext[$name]}" >>$json
+            elif [ ! -z "${PsecurityContext[default]}" ]; then
+                echo -e ",${PsecurityContext[default]}" >>$json
+            fi
         fi
-        set -u
```
k-rister, line 743: Isn't the `set -u` needed right before the above `fi`?

atheurer: It is needed, but right after the if

k-rister (approved):

PR #483, opened 2024-03-27
remotehosts.py endpoint

k-rister:


k-rister:
Just created for tracking purposes at this point.

k-rister:
This PR depends on https://github.com/perftool-incubator/crucible/pull/340 being merged and a new controller image being created which includes it.

k-rister:
Another PR that this PR is dependent on: https://github.com/perftool-incubator/toolbox/pull/76

k-rister:
I have a crucible-ci commit that is pending submission as a PR once this PR is complete that will add testing for this new endpoint:

https://github.com/k-rister/crucible-ci/commit/92b2710036c1196d6ecedb1d2194bc78830bd3c0

util/ci-run-file-creator.py
```diff
@@ -222,6 +222,35 @@ def update_userenvs():
...
+            endpoint = run_file["endpoints"][0]
+            if "userenv" in endpoint["settings"]:
+                if args.userenv == "default":
+                    log.info("Found existing %s global userenv but since requested userenv is default is is being removed" % (args.endpoint))
+                    del endpoint["settings"]["userenv"]
+                else:
+                    log.info("Updating existing %s global userenv" % (args.endpoint))
+                    endpoint["settings"]["userenv"] = args.userenv
+            else:
+                if args.userenv == "default":
```
rafaelfolco, line 235: if there is no "userenv" defined, how this could be != default ?

util/ci-run-file-creator.py
```diff
@@ -159,7 +159,7 @@ def update_endpoint_sub_type():
     run_file["tags"]["endpoint-sub-type"] = args.endpoint_sub_type
 
     match args.endpoint:
-        case "remotehost":
+        case "remotehost"|"remotehosts":
```
rafaelfolco, line 162: could your please refresh me on the needs of both?

util/ci-run-file-creator.py
```diff
@@ -222,6 +222,35 @@ def update_userenvs():
...
+                    del endpoint["settings"]["userenv"]
+                else:
+                    log.info("Updating existing %s global userenv" % (args.endpoint))
+                    endpoint["settings"]["userenv"] = args.userenv
+            else:
+                if args.userenv == "default":
+                    log.info("No %s global userenv to update since requested userenv is default" % (args.endpoint))
+                else:
+                    log.info("Creating %s global userenv since requested userenv is not default" % (args.endpoint))
+                    endpoint["settings"]["userenv"] = args.userenv
```
rafaelfolco, line 239: if not default and not defined in the json, where this args.userenv comes from ?

util/ci-run-file-creator.py
```diff
@@ -286,6 +333,15 @@ def update_host():
                     endpoint["host"] = args.host
                 else:
                     log.debug("No host to update for %s endpoint %d" % (args.endpoint, endpoint_idx))
+        case "remotehosts":
+            endpoint = run_file["endpoints"][0]
```
rafaelfolco, line 337: why endpoints index 0 ? Isn't this an array of endpoints ?

util/blockbreaker.py
```diff
@@ -130,6 +130,9 @@ def json_to_stream(json_obj, cfg, idx):
             endpoint_type = json_blk["type"]
             if not validate_schema(json_blk, "schema-" + endpoint_type + ".json"):
                 return None
+            if endpoint_type == "remotehosts":
+                stream = endpoint_type + ",run-file=" + args.json_file + ",endpoint-index=" + str(idx)
```
rafaelfolco, line 134: so the endpoint_type is not part of the run-file ? why still need to add it here ?

rafaelfolco (commented):
code looks good, just raised a few questions for clarity

util/blockbreaker.py
k-rister: This "stream" is being sent to `rickshaw-run` -- it uses the first part (the `endpoint_type`) to know which endpoint to invoke and then the rest is sent do the endpoint itself.

util/ci-run-file-creator.py
k-rister: Both `remotehost` and `remotehosts`?  This script is being used to convert templates into usable JSON run-files for all endpoints.  This particular function is currently only doing something meaningful for the `k8s` endpoint so both `remotehost` and `remotehosts` are lumped together in what amounts to a `noop` case.

util/ci-run-file-creator.py
k-rister: So a little bit of a cheat/assumption here.  Yes, the JSON is capable of hosting many endpoints in the endpoints array -- BUT, in CI only a single endpoint type is being processed at any one time AND the remotehosts endpoint does not support multiple invocations of itself in the same JSON.  So, by process of elimination the remotehosts endpoint must be at array index 0.

The reason the remotehosts endpoint does not support multiple invocations of itself comes down to 2 reasons:

1. It should not ever need multiple invocations.  Since it supports multiple remotes and each remote can have it's own set of endpoint settings there is no need to have multiple endpoints of the same remotehosts type.  If at some point we find a reason to need this, I view that as bug that needs to be fixed.
2. This is probably the more important of the two reasons.  The remotehosts endpoint dedups the list of remotes into a unique set that it uses to perform certain operations that should only be performed once per remote such as pulling container images, performing container image cache management, and determining where profiler engines should be run.  If multiple remotehosts endpoints where specified it would not be able to properly dedup the remotes across the multiple invocations and would then perform these operations multiple times on the same remotes.  This was one of the scalability problems detected with the current remotehost endpoint -- certain operations such as pulling container images would breakdown when being done too many times in parallel on the same remote.

util/ci-run-file-creator.py
k-rister: The invocation of the script by `crucible-ci`:

https://github.com/perftool-incubator/crucible-ci/blob/be052c3b92a2e15fb8ed2f5a8e13e542094e697d/.github/actions/integration-tests/run-ci-stage1#L534-L578

rafaelfolco (approved):

PR #486, opened 2024-04-22
Dev kmr4

k-rister:


k-rister:
@rafaelfolco

It's probably going to be hard to read/understand this, but the code in this PR will be "turned on" by a subsequent PR in crucible-ci that uses it.  That PR will be based on this work:

https://github.com/perftool-incubator/crucible-ci/compare/main...k-rister:crucible-ci:dev-kmr

rafaelfolco (approved):
clean code, very nice :)

PR #492, opened 2024-05-02
update remotehosts.py to formally name the threads it launches

k-rister:
- this is helpful because if Python throws an error message it will include the formal name of the thread which can be useful for tracking where the problem came from

### DIFF
--- a/endpoints/remotehosts/remotehosts.py
+++ b/endpoints/remotehosts/remotehosts.py
@@ -1288,10 +1288,10 @@ def image_pull_worker_thread(thread_id, work_queue, threads_rcs):
             remote = work_queue.get(block = False)
         except queue.Empty:
             thread_logger(thread_name, "Received a work queue empty exception")
-            continue
+            break
 
         if remote is None:
-            thread_logger(thread_name, "Received a null job")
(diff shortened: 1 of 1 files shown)


endpoints/remotehosts/remotehosts.py
```diff
@@ -101,7 +101,7 @@ def remote_connection(host, user):
                     validate_comment(msg)
                 else:
                     log.info(msg)
-                break
```
rafaelfolco, line 104: did this cause any side effects ?

rafaelfolco (approved):

endpoints/remotehosts/remotehosts.py
k-rister: It was causing the code to iterate through the loop for at least 2 iterations, even if it succeeded on the first.  This would result in the `connection.open()` being called at least twice no matter what.  It's not exactly clear to me what calling that a second time after it had succeeded the first time might mean, but I could see that it might have added some extra overhead.

It was also weird that it was logging that it took a 2nd attempt when there was no logged failed attempt.  This is what clued me into there being a problem.

PR #495, opened 2024-05-14
Dev kmr

k-rister:


atheurer (approved):
I think this looks like to me.  We still get the scp output (to the log file), just not stored in a var first, right?

k-rister:
> I think this looks like to me. We still get the scp output (to the log file), just not stored in a var first, right?

Yes, in fact we now get the scp output everytime whereas before we would only get the output if it failed.  Given that the output can yield valuable information about the timing of the steps in the connection and whether or not we are succesfully using the multiplexed session I think that always outputting it is a wise move.

PR #497, opened 2024-05-17
minor bugfix in re

atheurer:


endpoints/remotehosts/remotehosts.py
```diff
@@ -846,13 +846,14 @@ def is_ip(ip_address):
     Returns:
         True or False
     """
+    print("#checking to see if ('%s') is IP" % (ip_address))
```
k-rister, line 849: This should probably be something like this:

```
msg = "Checking to see if '%s' is an IP address" % (ip_address)
if args.validate:
    validate_comment(msg)
else:
    log.info(msg)
```

k-rister (changes requested):

endpoints/remotehosts/remotehosts.py
atheurer: Crap, I forgot I left that in.  Do you think this output is necessary/useful?

k-rister: Not really.

k-rister (approved):

PR #498, opened 2024-05-22
extract all shared/reusable code from remotehosts.py and create endpo…

k-rister:
…ints.py containing it

### DIFF
--- a/endpoints/endpoints.py
+++ b/endpoints/endpoints.py
@@ -1,0 +1,1426 @@
+import calendar
+from fabric import Connection
+from invoke import run
+import json
+import logging
+import os
+from pathlib import Path
+import re
+import sys
+import time
+
+TOOLBOX_HOME = os.environ.get('TOOLBOX_HOME')
+if TOOLBOX_HOME is None:
+    print("This script requires libraries that are provided by the toolbox project.")
+    print("Toolbox can be acquired from https://github.com/perftool-incubator/toolbox and")
+    print("then use 'export TOOLBOX_HOME=/path/to/toolbox' so that it can be located.")
+    exit(1)
+else:
+    p = Path(TOOLBOX_HOME) / 'python'
+    if not p.exists() or not p.is_dir():
+        print("ERROR: <TOOLBOX_HOME>/python ('%s') does not exist!" % (p))
+        exit(2)
+    sys.path.append(str(p))
+from toolbox.json import *
+
+ROADBLOCK_HOME = os.environ.get('ROADBLOCK_HOME')
+if ROADBLOCK_HOME is None:
+    print("This script requires libraries that are provided by the roadblock project.")
+    print("Roadblock can be acquired from https://github.com/perftool-incubator/roadblock and")
+    print("then use 'export ROADBLOCK_HOME=/path/to/roadblock so that it can be located.")
+    exit(1)
+else:
+    p = Path(ROADBLOCK_HOME) / 'roadblock.py'
+    if not p.exists() or not p.is_file():
+        print("ERROR: <ROADBLOCK_HOME>/roadblock.py ('%s') does not exist!" % (p))
+        exit(2)
+    sys.path.append(str(Path(ROADBLOCK_HOME)))
+from roadblock import roadblock
+
+roadblock_exits = {
+    "success": 0,
+    "timeout": 3,
+    "abort": 4,
+    "input": 2
+}
+
+log = logging.getLogger(__file__)
+
+def run_remote(connection, command, validate = False, debug = False):
+    """
+    Run a command on a remote server using an existing Fabric connection
+
+    Args:
+        connection (Fabric Connection): The connection to use to run the command remotely
+        command (str): The command to run
+        validate (bool): Is the function being called from validation mode (which means that logging cannot be used)
+        debug (bool): Is debug output enabled during validation mode
+
+    Globals:
+        None
+
+    Returns:
+        a Fabric run result
+    """
+    debug_msg = "on remote '%s' as '%s' running command '%s'" % (connection.host, connection.user, command)
+    if validate:
+        if debug:
+            validate_debug(debug_msg)
+    else:
+        log.debug(debug_msg, stacklevel = 2)
+
+    return connection.run(command, hide = True, warn = True)
+
+def run_local(command, validate = False, debug = False):
+    """
+    Run a command on the local machine using Invoke
+
+    Args:
+        command (str): The command to run on the local system
+        validate (bool): Is the function being called from validation mode (which means that logging cannot be used)
+        debug (bool): Is debug output enabled during validation mode
+
+    Globals:
+        None
+
+    Returns:
+        an Invoke run result
+    """
+    debug_msg = "running local command '%s'" % (command)
+    if validate:
+        if debug:
+            validate_debug(debug_msg)
+    else:
+        log.debug(debug_msg, stacklevel = 2)
+
+    return run(command, hide = True, warn = True)
+
+def get_result_log_level(result):
+    """
+    Determine the log level to return based on the provided result's return code.
+
+    Args:
+        result (Fabric/Invoke result): From a Fabric/Invoke run command
+
+    Globals:
+        None
+
+    Returns:
+        str: The appropriate log level to use to describe the result
+    """
+    if result.exited != 0:
+        return "error"
+    else:
+        return "info"
+
+def remote_connection(host, user, validate = False):
+    """
+    Create a Fabric connection and open it
+
+    Args:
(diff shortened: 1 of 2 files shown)


endpoints/endpoints.py
```diff
@@ -0,0 +1,1427 @@
...
+        exit(2)
+    sys.path.append(str(Path(ROADBLOCK_HOME)))
+from roadblock import roadblock
+
+roadblock_exits = {
+    "success": 0,
+    "timeout": 3,
+    "abort": 4,
+    "input": 2
+}
```
rafaelfolco, line 45: wondering if these numbers need to be updated if they change on roadblock's side

endpoints/endpoints.py
```diff
@@ -0,0 +1,1427 @@
...
+        connection (Fabric Connection): The connection to use to run the command remotely
+        command (str): The command to run
+        validate (bool): Is the function being called from validation mode (which means that logging cannot be used)
+        debug (bool): Is debug output enabled during validation mode
+
+    Globals:
+        None
+
+    Returns:
+        a Fabric run result
```
rafaelfolco, line 63: I like the docstring for the functions. Fabric result is int ? We could adopt (type) for the return as well.

endpoints/endpoints.py
```diff
@@ -0,0 +1,1427 @@
...
+            break
+        except (ssh_exception.AuthenticationException, ssh_exception.NoValidConnectionsError) as e:
+            msg = "Failed to connect to remote '%s' as user '%s' on attempt %d due to '%s'" % (host, user, attempt, str(e))
+            if validate:
+                validate_comment(msg)
+            else:
+                log.warning(msg)
+
+            if attempt == attempts:
+                msg = "Failed to connect to remote '%s' as user '%s' and maximum number of attempts (%d) has been exceeded.  Reraising exception '%s'" % (host, user, attempts, str(e))
```
rafaelfolco, line 153: Speaking of code standards and best practices.... these long lines could be 
msg = ("my 1st line is too long"
             "so I decided to break it"
             "into more lines")

endpoints/endpoints.py
```diff
@@ -0,0 +1,1427 @@
...
+                if validate:
+                    validate_error(msg)
+                else:
+                    log.error(msg)
+
```
rafaelfolco, line 158: nit: any reason for these blank lines ?

endpoints/endpoints.py
rafaelfolco: https://peps.python.org/pep-0008/#maximum-line-length

endpoints/endpoints.py
```diff
@@ -0,0 +1,1427 @@
...
+    Globals:
+        None
+
+    Returns:
+        str: a representation of the object that can be serialized into a json property value
+    """
+    try:
+        return(obj.to_dictionary())
+    except AttributeError:
+        return(repr(obj))
```
rafaelfolco, line 282: nit: return xyz

rafaelfolco (dismissed):
lgtm for a quick pass - few comments on code style

endpoints/endpoints.py
k-rister: They would.  Now, in theory this is the kind of thing that I would never change on the roadblock side in order to maintain a stable "interface".

It does remind me that in the newer model of using roadblock.py as a module that I could potentially refer to these codes from the module that is being imported rather than replicated.  I'll have to look into that.

endpoints/endpoints.py
k-rister: It is not an int, it is an object with multiple properties:

https://docs.pyinvoke.org/en/latest/api/runners.html#invoke.runners.Result

endpoints/endpoints.py
k-rister: So I'm not opposed to doing something about the line length, but I think the solution needs to be carefully chosen.  I genuinely hate most of the solutions I have seen for this because of (my perception) the ambiguity that they create for the formatting.  In the example that you provided:

```
msg = ("my 1st line is too long"
"so I decided to break it"
"into more lines")
```

Does that automatically put spaces between `long` and `so` and `it` and `into`?  What if you don't want spaces there?  What if you want multiple?

endpoints/endpoints.py
k-rister: I like blank lines and I cannot lie?

endpoints/endpoints.py
k-rister: Yep, oops.

endpoints/endpoints.py
k-rister: I fixed this because there was also an indentation error.

endpoints/endpoints.py
k-rister: I fixed this also.

endpoints/endpoints.py
```diff
@@ -0,0 +1,1426 @@
...
+    Args:
+        msg (str): The message to log
+
+    Globals:
+        None
+
+    Returns:
+        None
+    """
+    return print("# DEBUG: " + msg)
```
rafaelfolco, line 221: refresh me on why we can't merge these validate_xxx into one and add an arg... validate_log(msg, level)

endpoints/endpoints.py
```diff
@@ -0,0 +1,1426 @@
...
+    """
+    log.info("Checking to see if '%s' is an IP address" % (ip_address))
+
+    # check for IPv4
+    m = re.search(r"[1-9][0-9]{0,2}\.[1-9][0-9]{0,2}\.[1-9][0-9]{0,2}\.[1-9][0-9]{0,2}", ip_address)
+    if m:
+        return True
+
+    # check for IPv6
+    m = re.search(r"[0-9a-fA-F]{1,4}:[0-9a-fA-F]{1,4}:[0-9a-fA-F]{1,4}:[0-9a-fA-F]{1,4}:[0-9a-fA-F]{1,4}:[0-9a-fA-F]{1,4}:[0-9a-fA-F]{1,4}:[0-9a-fA-F]", ip_address)
```
rafaelfolco, line 381: If we change here or in the schema, we're out of sync... not sure how to solve this though

endpoints/endpoints.py
```diff
@@ -0,0 +1,1426 @@
...
+
+    rc = do_roadblock(roadblock_id = roadblock_id,
+                      follower_id = endpoint_label,
+                      label = "endpoint-deploy-begin",
+                      timeout = endpoint_deploy_timeout,
+                      messages = new_followers_msg_file,
+                      max_attempts = max_roadblock_attempts,
+                      redis_password = roadblock_password,
+                      msgs_dir = roadblock_messages_dir)
+    if rc != 0:
```
rafaelfolco, line 887: these blocks are identical, do you keep them separate for readability reasons ?

rafaelfolco (approved):
superficial review, lgtm

endpoints/endpoints.py
```diff
@@ -0,0 +1,1426 @@
...
+                if len(split_line) > 1:
+                    split_line = split_line[1]
+                    if len(split_line) > 1:
+                        controller_ip = split_line[0]
+                        break
+        if controller_ip is None or not is_ip(controller_ip):
+            raise ValueError("Failed to map localhost to loopback IP address (got '%s')" % controller_ip)
+        else:
+            host = controller_ip
+            controller_ip = None
```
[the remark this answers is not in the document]
k-rister: It is "resetting" the values going into the next block of code to put them in a similar state to how the function starts.

endpoints/endpoints.py
k-rister: Yes, the duplication is about simplicty and readability.  We actually had a loop based implementation written in Bash at one point and it was particularly confusing due to the different exceptions that were having to be made for the different tiers.

This code is essentially duplicated in 3 different places (tiers):

1. `rickshaw-run`
2. the endpoint(s)
3. the engine

Each tier has different exceptions and activities that happen in between the roadblocks.  When each one was implemented in a loop we were forced to look at 3 different loops with varying exceptions and visualizing and understanding it was very, very difficult.  So one of the first things I did was unroll the loops so that now you can search for a specific roadblock (ex. `collect-sysinfo-end`) in each location and very easily see what each "tier" of the stack is doing compared to the others.

Additionally, if we want to add additional stages you just go to the proper place in the flow and add more code (probably via copy/paste) because it is so much easier to determine what the proper place is.

endpoints/endpoints.py
k-rister: FWIW, the logging library uses `<obj>.info(...)`, `<obj>.error(...)`, `<obj>.debug(...)` instead of `<obj>.log(..., level)` for some reason....

PR #499, opened 2024-05-28
bugfix: make sure rickshaw-run knows how to handle a waiting abort si…

k-rister:
…gnal

### DIFF
--- a/rickshaw-run
+++ b/rickshaw-run
@@ -114,6 +114,7 @@ my $roadblock_exit_success = 0;
 my $roadblock_exit_timeout = 3;
 my $roadblock_exit_abort = 4;
 my $roadblock_exit_input = 2;
+my $roadblock_exit_abort_waiting = 6;
 my $abort_via_roadblock = 0;
 my $workshop_base_cmd;
 my $workshop_force_builds;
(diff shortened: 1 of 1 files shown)


rafaelfolco (approved):
a quick description of what is the waiting abort would be much appreciated

k-rister:
> a quick description of what is the waiting abort would be much appreciated

So a waiting abort is happens when roadblock is operating in a specific mode (called `wait-for`) where one of the participants has been told to launch a subprocess and block the progression of the roadblock until the subprocess exits.  If the subprocess exits with a non-zero return code then roadblock considers that an error and rather than completing the roadblock successfully it aborts it.  For the roadblock side of things I decided to make roadblock itself exit with a `waiting-abort` rather than an `abort` so that it was very clearly communicating to the caller why it was aborting -- but the caller may not actually care why.  That is what is happening here with `rickshaw-run` -- it does not care why there is an abort, it just needs to know that an abort is happening.  So this patch is adding minimal logic to make `rickshaw-run` handle `waiting-abort` the same as `abort`.

Right now Crucible is only using `wait-for` in a very few limited scenarios and so this happening is fairly rare, which is how/why this bug exists.  One scenario is the `send-data` phase since we really have no idea how much data an engine may be sending back and/or how fast/slow the connection is.  This practically never fails, so the bug never surfaced there.  The other scenario, which is what surfaced the bug, is that trafficgen uses `wait-for` to run the benchmark since a binary-search has an unbounded runtime.  So when trafficgen was exiting with an error it was causing a `waiting-abort`.

PR #502, opened 2024-06-10
Dev kmr

k-rister:


.github/workflows/crucible-merged.yaml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  call-core-crucible-ci_if-merged:
+    if: github.event.pull_request.merged == true
+    uses: perftool-incubator/crucible-ci/.github/workflows/core-crucible-ci.yaml@main
+    with:
+      ci_target: "rickshaw"
+      ci_target_branch: "master"
+      github_workspace: "$GITHUB_WORKSPACE"
+    secrets:
+      production_registry_auth: ${{ secrets.CRUCIBLE_PRODUCTION_ENGINES_REGISTRY_AUTH }}
```
rafaelfolco, line 26: ok you push to production when merged. 
does ci continue to use the same ci images or converge to production ones at some point ?

rafaelfolco (approved):

.github/workflows/crucible-merged.yaml
k-rister: They remain separate.

All CI runs wil use the CI repo (and corresponding token) to isolate images built with experimental code from production environments.  Once PRs are merged the jobs will be re-run with the production repo (and corresponding token) to populate the production repo with long-lived images to prevent users from having to build images themselves as much as possible.

The entire image building infrastructure will still be present and active on all controllers so that if a user makes local modifications that require updated images they can be built for them.

My current plan is for the CI built production images to have a very long life (52 weeks == 1 year) and to leave custom built images at the current 2 week life.

PR #510, opened 2024-07-12
Dev kmr2

k-rister:


endpoints/remotehosts/remotehosts.py
```diff
@@ -1985,8 +1987,38 @@ def remote_image_manager(thread_name, remote_name, connection, image_max_cache_s
...
+
+        thread_logger(thread_name, "images[quay]:\n%s" % (endpoints.dump_json(images["quay"])), remote_name = remote_name, log_prefix = log_prefix)
+    else:
+        thread_logger(thread_name, "Configuration information necessary to utilize the quay API to obtain image expiration timestamps is not available", remote_name = remote_name, log_prefix = log_prefix)
+
     image_expiration = endpoints.image_expiration_gmepoch()
-    thread_logger(thread_name, "Images created before %d will be considered expired" % (image_expiration), remote_name = remote_name, log_prefix = log_prefix)
+    thread_logger(thread_name, "Images evaludated by their expiration data will be considered expired if it is before %d" % (image_expiration), remote_name = remote_name, log_prefix = log_prefix)
+
+    expiration_weeks = int(settings["rickshaw"]["quay"]["image-expiration"].rstrip("w"))
```
rafaelfolco, line 2019: If I understand it right, this will be used ONLY if expiration cannot be retrieved from the registry, as a fallback.
What if the expiration is not defined in the rickshaw config for some reason ? Could not find a default value (previously 2w).

endpoints/remotehosts/remotehosts.py
```diff
@@ -2020,12 +2052,24 @@ def remote_image_manager(thread_name, remote_name, connection, image_max_cache_s
...
+
+        if image in images["quay"]:
+            if images["quay"][image]["end_ts"] < image_expiration:
+                thread_logger(thread_name, "Podman image '%s' has been evaluated based on it's expiration data and has expired, removing it from the image cache" % (image), remote_name = remote_name, log_prefix = log_prefix)
+                deletes["podman"].append(image)
+                deletes["rickshaw"].append(image)
+                remove_image(thread_name, remote_name, log_prefix, connection, image)
+            else:
+                thread_logger(thread_name, "Podman image '%s' has been evaluated based on it's expiration data and has not expired" % (image), remote_name = remote_name, log_prefix = log_prefix)
+        elif images["podman"][image]["created"] < image_created_expiration:
```
rafaelfolco, line 2064: Are you sure images["podman"][image] exists ? This is a 'elif' for  if image in images["quay"]:, so not having  if image in images["quay"] implies in having  if image in images["podman"] ?

rafaelfolco (approved):
lgtm, questions raised

endpoints/remotehosts/remotehosts.py
k-rister: At the moment there is a "loose" guarantee that the expiration is in the `rickshaw-settings.json` file -- meaning that the upstream config file has a value in it.  If a user, for some reason, deleted that value from the config file then this would blow up.  The `2w` default is the value that is in the upstream repo for this parameter in the `rickshaw-settings.json` file:

https://github.com/perftool-incubator/rickshaw/blob/01fbff69fd0a9f35e1e5caa952bbde09f36a52c2/rickshaw-settings.json#L26

The "real" solution that I intend on doing at some point is to create a JSON schema for the `rickshaw-settings.json` file -- just like we have for so many other JSON files -- and so there would be required parameters, such as this expiration setting, and data formats that would cause validation failures much earlier in the execution cycle if they are absent/incorrect.  I'm not exactly sure why we didn't do that from the beginning, it's likely an oversight on my part.

endpoints/remotehosts/remotehosts.py
k-rister: The loop is `for image in images["podman"].keys():`.

PR #521, opened 2024-07-18
add LoadBalancer support, yet another ingress service

HughNhan:
### Synopsis:
Add LoadBalancer (LB) support, yet another INGRESS service
### Description: 
The LB service is used in Nokia Core cluster where it has dedicated Gateway nodes for ingress/egress.
On baremetal, MetalLB is the LB control plane. To setup LB service, we need to install MetalLB, an IPAddressPool object . In the L2 mode where routes are resolved by ARP, we also install a L2Advertisement object. All those setup are outside Crucible.
In rickshaw, we detect Ingress topology when the workload server is on k8s, and the client is outside of k8s. In the past, for ingress we support nodePort by default. Now we have a choice between nodePort and LB. We select this choice by a new endpoint option "lbSvc=<lb_pool_name>", the "lb_pool_name" selects a IPAddressPool object.
### Code changes

1. Add an "if-else" for nodePort versus LB code path.
2. Add a function to create the LB service CR
3. Add a label to client and server pod so that LB will select the app

### Test:

- Both nodePort and the new LB are working.

### DIFF
--- a/endpoints/k8s/k8s
+++ b/endpoints/k8s/k8s
@@ -57,6 +57,7 @@ project_name="crucible-rickshaw"
 hypervisor_host="none"
 unique_project="0"
 hostNetwork="0"
+lbSvc=""
 hugepage="0"
 osruntime[default]="pod"
 runtimeClassName=""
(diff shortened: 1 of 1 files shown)


endpoints/k8s/k8s
```diff
@@ -543,6 +629,13 @@ function build_pod_spec() {
     echo "  ,\"kind\": \"Pod\"" >>$json
     echo "  ,\"metadata\": {" >>$json
     echo "    \"name\": \"$pod_prefix-$name\"," >>$json
+
+    if [ "$type" == "cs" ]; then
+        # server pod needs a label for loadBalancer svc to select.
+        echo "    \"labels\": {"  >>$json
+        echo "        \"app\": \"$pod_prefix-$name\""  >>$json
+        echo "    }," >>$json
+    fi
```
k-rister, line 638: Should this be always done whether the loadBalancer svc is setup/active or not?  Or should it only be done when it is active?

HughNhan: To make it simpler,  just have a label always.

k-rister (approved):

PR #522, opened 2024-07-19
Add support for podman-settings

### LINKED ISSUE #516: Add support for podman-settings in remotehosts.py, including a new setting "device"
An example of this feature:

```
{
  "tags": {
    "topology": "none"
  },
  "endpoints": [
    {
      "type": "remotehosts",
      "remotes": [
        {
          "engines": [ { "role": "client", "ids": "1" } ],
          "config": {
            "host": "localhost",
            "settings": {
              "controller-ip-address": "10.26.8.21",
              "userenv": "rhel-ai",
              "osruntime": "podman",
              "podman-settings": {
                "device": "nvidia.com/gpu=all"
              },
              "cpu-partitioning": false
            }
          }
        }
      ]
    }
  ],
  "run-params": { "num-samples": 1, "max-sample-failures": 1, "test-order": "r" },
  "tool-params": [
    { "tool": "sysstat",
      "params": [
        { "arg": "subtools", "val": "mpstat,sar,iostat" },
        { "arg": "interval", "val": "15" }
      ]
    },
    { "tool": "procstat",
      "params": [
        { "arg": "interval", "val": "15" }
      ]
    }
  ],
  "benchmarks": [
    {
      "name": "ilab",
      "ids": "1",
      "mv-params": {
        "sets": [
          {
            "params": [
	      { "arg": "cpu-offload-optimizer", "vals": [ "1" ] },
	      { "arg": "cpu-offload-pin-memory", "vals": [ "1" ] },
	      { "arg": "nnodes", "vals": [ "1" ] },
	      { "arg": "model", "vals": [ "/home/models/granite-7b-lab/" ] },
	      { "arg": "data-path", "vals": [ "/home/data/jun12-phase05.jsonl" ] },
	      { "arg": "nproc-per-node", "vals": [ "2" ] }
	    ]
          }
	]
      }
    }
  ]
}
```

In this case, the device option results in the following used for the podman run command: `--device nvidia.com/gpu=all`

### LINKED ISSUE #518: Add support for podman shared memory
Needs to include size, like "10.00gb", which results in a podman run option of `--shm-size=10.00gb`

atheurer:
-allows the use of podman features:
 -device
 -shm-size

 ...which are required for GPU containers

### DIFF
--- a/schema/remotehosts.json
+++ b/schema/remotehosts.json
@@ -140,7 +140,21 @@
                 "userenv": {
                     "type": "string",
                     "minLength": 1
-                }
+                },
+		"podman-settings": {
+                    "type": "object",
+		    "properties": {
+                        "device": {
+                            "type": "string",
+                            "minLength": 1
+                        },
+                        "shm-size": {
+                            "type": "string",
+                            "minLength": 1
+                        }
+                    },
+                    "minLength": 1
+		}
             },
             "additionalProperties": false
         },
--- a/endpoints/remotehosts/remotehosts.py
+++ b/endpoints/remotehosts/remotehosts.py
@@ -51,6 +51,7 @@ endpoint_defaults = {
     "numa-node": None,
     "osruntime": "podman",
     "user": "root",
+    "podman-settings": {},
     "maximum-worker-threads-count": 250
 }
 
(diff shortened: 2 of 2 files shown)


schema/remotehosts.json
```diff
@@ -140,7 +140,21 @@
...
+                    "type": "object",
+		    "properties": {
+                        "device": {
+                            "type": "string",
+                            "minLength": 1
+                        },
+                        "shm-size": {
+                            "type": "string",
+                            "minLength": 1
+                        }
```
k-rister, line 154: Would it be possible to make the validations of these properties any more accurate or are they really completely arbitrary strings?

k-rister (approved):
I asked one question, but otherwise it looks good to me.

schema/remotehosts.json
atheurer: I think over time it would be a good idea to have these match something like a regex.  I am not yet familiar with how to do that in the json validator.  Or maybe better suited for endpoint validation?

PR #549, opened 2024-08-30
Fix `userenv.json` loading

zeeke:
Userenv directory are in the for `./userenvs/<name>/userenv.json`

Change the loading path to avoid errors like:
```
Working on stream8 userenv
Sourcing container image for userenv 'stream8' and benchmark/tool 'iperf'; this may take a few minutes
ERROR: Could not load userenv JSON file for 'stream8' due to non-zero return code 8.  Are you sure this is a supported userenv?
Perl exited with active threads:
        0 running and unjoined
        1 finished and unjoined
        0 running and detached
Skipping run post-processing due to error(s) [rc=2]
```

### DIFF
--- a/rickshaw-run
+++ b/rickshaw-run
@@ -898,9 +898,10 @@ sub source_container_image {
     my @workshop_args;
     my $userenv_arg;
     my $count = 0;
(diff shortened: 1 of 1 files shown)


zeeke:
I didn't get this

https://github.com/perftool-incubator/rickshaw/blob/master/userenvs/README#L3

sorry for the noise, closing

PR #556, opened 2024-10-23
[bugfix] Change Bash regex for engine labels

rafaelfolco:
It is now capable of capturing a profiler name with dashes in the middle of the string - e.g. rt-trace-bpf.

### DIFF
--- a/engine/engine-script-library
+++ b/engine/engine-script-library
@@ -305,14 +305,25 @@ function validate_core_env() {
     if [ -z "$cs_label" ]; then
         exit_error "The client/server label (--cs-label) was not defined"
     fi
-    # profiler-remotehost-1-sysstat-1
-    regex1='^profiler-\w+-\d+-\w+-\d+$'
-    # client-1
-    regex2='^\w+-\d+$'
-    if echo $cs_label | grep -q -P $regex1; then
-        export tool_name=`echo $cs_label | awk -F- '{print $4}'`
-        echo "engine-label \"$cs_label\" is valid and runs this tool:  $tool_name"
-    elif echo $cs_label | grep -q -P $regex2; then
+
+    # (rfolco): regex 101
+    #     ^profiler- ensures the string starts with profiler-
+    #     ([a-zA-Z0-9]+) captures one or more lowercase/uppercase letters or numbers like k8s
+    #     ([0-9]+) captures one or more digits for the third part
+    #     ([a-zA-Z-]+) captures one or more lowercase/uppercase letters or dashes for the fourth part
+    #     ([0-9]+)$ ensures the string ends with a number after the last part
+
+    # e.g. profiler-remotehost-1-sysstat-1,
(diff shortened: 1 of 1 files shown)


engine/engine-script-library
```diff
@@ -305,14 +305,20 @@ function validate_core_env() {
...
+    #     ([a-z]+) captures one or more lowercase letters for the second part.
+    #     ([0-9]+) captures one or more digits for the third part.
+    #     ([a-z_-]+) captures one or more lowercase letters, underscores, or dashes for the fourth part.
+    #     ([0-9]+)$ ensures the string ends with a number after the last part.
+    #     e.g. profiler-remotehost-1-sysstat-1 or profiler-remotehost-1-rt-trace-bpf-1
+    regex1='^profiler-([a-z]+)-([0-9]+)-([a-z_-]+)-([0-9]+)$'
+    # client-1 or server-1
+    regex2='^([a-z]+)-([0-9]+)$'
+    if [[ $cs_label =~ $regex1 ]]; then
         export tool_name=`echo $cs_label | awk -F- '{print $4}'`
```
k-rister, line 319: Doesn't this need to be rewritten as well?

engine/engine-script-library
```diff
@@ -305,14 +305,20 @@ function validate_core_env() {
...
         exit_error "The client/server label (--cs-label) was not defined"
     fi
-    # profiler-remotehost-1-sysstat-1
-    regex1='^profiler-\w+-\d+-\w+-\d+$'
-    # client-1
-    regex2='^\w+-\d+$'
-    if echo $cs_label | grep -q -P $regex1; then
+    # (rfolco): regex 101
+    #     ^profiler- ensures the string starts with profiler-.
+    #     ([a-z]+) captures one or more lowercase letters for the second part.
```
k-rister, line 310: We should probably do `a-zA-Z` everywhere you have `a-z` now so that capital letters don't break it.

engine/engine-script-library
```diff
@@ -305,14 +305,20 @@ function validate_core_env() {
...
+    #     ([0-9]+) captures one or more digits for the third part.
+    #     ([a-z_-]+) captures one or more lowercase letters, underscores, or dashes for the fourth part.
```
k-rister, line 312: `a-zA-Z`

k-rister (changes requested):

engine/engine-script-library
rafaelfolco: ack

k-rister: Can you update the comments also?

engine/engine-script-library
k-rister: Can you update the comments also?

k-rister (changes requested):

k-rister (dismissed):

rafaelfolco:
Something is wrong with the new regex/logic. Need more debugging...

k-rister (approved):

PR #557, opened 2024-10-23
Only trim begin/end w/ pri-metric in pri-period

atheurer:


k-rister (commented):
Can you explain what/why you are making these changes?  It is not clear to me.

Also, I see a couple print statements -- are you really wanting to leave those in?  It seems like this could add a lot of output if it's per metri (or period).

atheurer:
The existing code was throwing an error for bench-ilab, where extra periods are created which don't have a primary-metric in that period, but a primary metric did exist (in another period).  When these metric files for the non-primary-period were processed, it failed because the check for earliest_begin and latest_end would fail.  However, those are not needed for non-primary periods

k-rister (approved):

PR #562, opened 2024-10-30
Dev kmr

k-rister:


rickshaw-run
```diff
@@ -1981,6 +1997,31 @@ sub load_settings_info() {
...
+        printf "load_settings_info(): loaded registry repository: %s\n", $run{'reg-repo'};
+    }
+
+    ($rc, $tmp_settings_value) = get_json_setting("engines.public.push-token", $registries_settings);
+    if ($rc != 0) {
+        print "load_settings_info(): failed to load public engines push-token\n";
+        exit 1;
+    } else {
+        $run{'reg-auth'} = $tmp_settings_value;
+        printf "load_settings_info(): loaded registry authorization token: %s\n", $run{'reg-auth'};
```
rafaelfolco, line 2016: this kind of info is more a debug level log, do we really want to print all these??

rafaelfolco (approved):
lgtm

rickshaw-run
k-rister: I think that I want it in while we go through these transitions to aide in debugging anything that might go wrong, but I may eventually decide to pull it back out for the reasons you state.

PR #563, opened 2024-10-31
Calculating begin/end needed further refinement

atheurer:
- See code comments for details


k-rister (approved):
I guess I have one curiosity, why are we only finding these problems now?  What is different about what you are doing?

PR #567, opened 2024-11-10
support cdmv8

atheurer:
- with backward compat with cdmv7
- cdmv8 will be introduced via CommonDataModel after this PR


k-rister (changes requested):
Something is broken, massive CI failures.

rickshaw-index
```diff
@@ -339,17 +404,29 @@ sub write_queued_es_docs {
 sub wait_for_metric_descs {
     my @terms = @_;
     my $attempts = 1;
-    my $max_attempts = 20;
+    my $max_attempts = 8;
```
k-rister, line 407: Are we sure about this change?  This seems like it might be a highly variable thing based on run size, no?

rickshaw-index
```diff
@@ -462,11 +539,12 @@ sub index_metrics {
...
-        if (! exists $$this_metr{'id'}) {
-            $$this_metr{'id'} = Data::UUID->new->create_str();
-            debug_log(sprintf "Adding persistent metric ID %s\n", $$this_metr{'id'});
-            $update_metric_file++;
-        }
+        #if (! exists $$this_metr{'id'}) {
+            #$$this_metr{'id'} = Data::UUID->new->create_str();
+            #debug_log(sprintf "Adding persistent metric ID %s\n", $$this_metr{'id'});
+            #$update_metric_file++;
+        #}
```
k-rister, line 546: Do we really need to keep this around?

k-rister (approved):

PR #570, opened 2024-11-22
related bugfixes

atheurer:


atheurer:
This includes some corrections for missing breakouts

rickshaw-index
```diff
@@ -459,7 +459,7 @@ sub index_metrics {
 
     my $dir =  pushd($metr_dir);
     my %eng_env_vars;
-    if ($cdm{'ver'} eq 'v7dev') {
```
atheurer, line 462: @HughNhan this is the fix that brings back the missing breakouts

k-rister (approved):

PR #571, opened 2024-11-24
restore deprecated properties to run.json schema

k-rister:
- even though these properties are no longer used they ar necessary for processing/indexing existing data

### DIFF
--- a/schema/run.json
+++ b/schema/run.json
@@ -42,6 +42,9 @@
       "type": "string",
       "pattern": "^.+$"
     },
+    "dest-image-url": {
+      "type": "string"
+    },
     "engine-dir": {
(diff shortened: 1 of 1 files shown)


k-rister:
@HughNhan Can you confirm that this patch fixes your indexing problem?

k-rister:
Assuming that this passes CI (I just had to make a typo bugfix), if @HughNhan says it's good for him then the `DO NOT MERGE` label no longer applies.

HughNhan:
It works. I can index the problematic run that the previous attempt before the fix would error out.

HughNhan (approved):
LGTM. Work great for me (WGFM)

PR #585, opened 2025-01-13
Add correct rhel-ai stage and offical userenvs

atheurer:


k-rister:
I think we are are going to have to do some cleanup work to make this all function properly.  Basically we are going to need to come up with a better solution than this:

https://github.com/perftool-incubator/crucible-ci/blob/main/.github/actions/get-userenvs/generate-userenv-list.sh#L8

If we are going to embed userenv's like this in the upstream rickshaw project rather than in a decicated userenv library.

k-rister:
First step: https://github.com/perftool-incubator/rickshaw/pull/586

k-rister:
@atheurer I have this rolling again...

k-rister (approved):

PR #595, opened 2025-01-31
update rhel ai userenvs

atheurer:
-adds accelerator type to name (nvidia, amd)
-adds bootc or ilab to name
-adds update-policy=never


k-rister (approved):
It occurred to me very recently while reading some discussion about these changes that I need to tweak the workflows and when I do so a PR like this wouldn't even run CI.  Just an optimization for me to add going forward, and it would be easiest to do if we stick the `rhel-ai` prefix on all these userenvs.

Issue #603, opened 2025-03-17
Change Default Redis Password


drewrip:
Currently in Rickshaw we [use a default password](https://github.com/perftool-incubator/rickshaw/blob/4192a97c19fdb32894f7261b95d13c48fc324ad0/rickshaw-run#L68) to authenticate Redis for use in Roadblock. This is not currently a security issue since we run Crucible benchmarks in clusters not publicly accessible. To future proof this however we should make sure that access to the Redis container is secure during the duration of the test when running in a publicly accessible cluster. I see two solutions to this:

1. Forward all the Redis traffic through the SSH, which already authenticated
2. Let the leader pick a random secure password and propagate the password to the followers via a previously established and secure SSH connection

From a user's perspective I don't think this is a DB that is particularly helpful for them to access, so neither of my proposed solutions include allowing the user to specify the password.

drewrip:
I created #604 as a proof of concept for solution (2). My hope with both suggested solutions were things that were somewhat easy to integrate and wouldn't add much latency. So essentially both just piggyback on the authenticated SSH connection. I wasn't entirely sure how to integrate (1), and I think that solution introduces a number of other failure points as well, so just creating a randomized password seems like a good start. Let me know what you think.

k-rister:
I think there is one bug you are going to have to fix for this to work and pass CI.  It looks like the `k8s` endpoint has the password hard coded which this is going to break:

```
$ git grep flubber
endpoints/k8s/k8s:        echo "            \"value\": \"flubber\"" >>$json
rickshaw-run:my $redis_passwd = "flubber"; # TODO: make this cmdline setting
```

k-rister:
FWIW, I think this is a great way to get some experience doing some behind the scenes dev work.  It's touching the endpoints, engine, and `rickshaw-run` but it is not a overly large problem to solve.

I agree that getting the password fixed up is a good first step and it's easy to do.  The SSH key thing is certainly doable, but it is more involved and probably better to do that in a second phase.

drewrip:
Hmm ok. Once I reopen the PR, I'll dig into testing it more thoroughly. I was slightly suprised this was confined to one simple variables. Also you asked I create a branch against the upstream repo, I don't think I have permission to push a new branch to the upstream repo.

k-rister:
I just sent you an invitation to our developers group which should give you write access.  I thought I had already done this, sorry for the mixup on my part.

drewrip:
[...]

PR #604, opened 2025-03-17
randomly generate common redis password

drewrip:
This implements solution (2) mentioned in #603. This will likely need some cleanup and testing as I am a perl novice.

The idea here is that we read 4096 bytes from /dev/urandom, then sha256 hash those bytes to get our random password. This password then replaces the previous hardcoded string.

### DIFF
--- a/rickshaw-run
+++ b/rickshaw-run
@@ -23,6 +23,7 @@ use threads;
 use threads::shared;
 use Thread::Queue;
 use Thread::Semaphore;
+use Digest::SHA qw(sha256_hex);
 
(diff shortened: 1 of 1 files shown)


k-rister:
@drewrip I haven't looked at this yet, but I am going to close it.  Can you resubmit it by creating a branch that you push to the upstream repository and then open the PR from that branch into master?  The way GitHub stuff works we need the PR to be from one branch to another within the upstream repository so that the resulting CI jobs can access the secrets we have stored for providing authentication access to [quay.io](quay.io) and a few other things.

PR #607, opened 2025-03-18
Add Support for Providing Roadblock Password in CLI Arg

drewrip:
This PR makes the roadblock password configurable via a cli argument: `--roadblock-password`. If none is provide it sets the normal default. This will go along with a corresponding PR to the Crucible repo towards accomplishing #603.

### DIFF
--- a/rickshaw-run
+++ b/rickshaw-run
@@ -1667,7 +1667,7 @@ sub process_cmdline() {
             $run{$p} = $defaults{$p};
         }
     }
-    my $base_rb_leader_cmd = "--role=leader --redis-server=localhost --redis-password=" . $run{'roadblock-password'};
+    $base_rb_leader_cmd = "--role=leader --redis-server=localhost --redis-password=" . $run{'roadblock-password'};
 }
 
 sub validate_controller_env() {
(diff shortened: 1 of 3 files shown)


rickshaw-run
```diff
@@ -1667,6 +1667,7 @@ sub process_cmdline() {
             $run{$p} = $defaults{$p};
         }
     }
+    my $base_rb_leader_cmd = "--role=leader --redis-server=localhost --redis-password=" . $run{'roadblock-password'};
```
k-rister, line 1670: This doesn't appear to be working and I'm wondering if the usage of `my` here is the reason why.  Does this mess with the variable scope?

@atheurer thoughts?

k-rister: > Ok I think this should be fixed with [this commit](https://github.com/perftool-incubator/rickshaw/pull/607/commits/0b9981732b5f3a3f218cd8221681b8027c368db9). Essentially just remove the `my` keywork. This should now reassign `base_rb_leader_cmd`, instead of creating a new local copy in that scope. As you can tell, I'm a perl novice

I think this is a pretty easy mistake to make.  I'm certainly no perl ninja but I've written a fair amount and I wasn't 100% sure that this was the problem until I did my example I just posted.

k-rister:
There is another lingering problem:

image

The rickshaw-run JSON schema needs to have the new parameter `roadblock-password` added to it.

drewrip:
Oh cool, I didn't realize there was validation going on. I added `roadblock-password` to the list.

k-rister (approved):

PR #609, opened 2025-03-21
remove the concept of multiple attempts for a specific roadblock

k-rister:
- this code was introduced at a time when we had some reasons to doubt that roadblock was behaving properly and is no longer needed

  - bugs were indeed fixed in roadblock that were occasionally causing problems

  - networking problems in OCP where causing packet loss which was breaking roadblock.  thanks in large part to roadblock those problems were identified and have been resolved

### DIFF
--- a/endpoints/base
+++ b/endpoints/base
@@ -14,7 +14,6 @@ rb_exit_success=0
 rb_exit_timeout=3
 rb_exit_abort=4
 rb_exit_input=2
-max_rb_attempts=""
 cs_rb_opts=""
 do_validate="0"
 num_clients=0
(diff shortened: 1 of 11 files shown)


endpoints/base
```diff
@@ -146,7 +145,7 @@ function exit_error() {
 function process_opts() {
     local longopts="validate,endpoint-opts:,endpoint-label:,run-id:,base-run-dir:,image:"
     local longopts+=",roadblock-server:,roadblock-passwd:,roadblock-id:,osruntime:,max-sample-failures:"
-    local longopts+=",max-rb-attempts:,endpoint-deploy-timeout:,engine-script-start-timeout:"
+    local longopts+=",,endpoint-deploy-timeout:,engine-script-start-timeout:"
```
atheurer, line 148: did you really mean to leave two commas here?

atheurer (dismissed):

endpoints/base
k-rister: No, good catch.

atheurer (approved):

PR #611, opened 2025-04-07
Cdmv9 multi opensearch

atheurer:
this introduced cdmv9, but cdmv8 is still the default version for now, [added] only because crucible does not use this script yet.  This PR supports cdm versions 7, 8, and 9.  Moving to v9 as the default will be accomplished in the crucible project

### DIFF
--- a/rickshaw-gen-docs
+++ b/rickshaw-gen-docs
@@ -5,8 +5,9 @@
 # Author: Andrew Theurer
 #
 # Rickshaw will run a benhcmark for you.  Please see README.md for instructions.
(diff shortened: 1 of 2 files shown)


atheurer:
Correction on the default version.  It is actually cdmv9, but since crucible does not use this script yet, that default does not affect the CDM version generated when submitting crucible runs.  Users can you this script to experiment with document generation (but skipping indexing) via rickshaw, then using add-run.js in CDM project to index.

k-rister (approved):

PR #627, opened 2025-06-04
Fix IP address validation regex

HughNhan:
The old regex has some deficiencies.  Both will not work with IPv6. Maybe we should use the python built-in function.

| IP Address | Original Regex | New Regex | Actual Validity |
|------------|-------------|--------------|-----------------|
| `192.168.1.1` | ✅ Match | ✅ Match | ✅ Valid |
| `192.168.0.1` | ❌ No match | ✅ Match | ✅ Valid |
| `10.0.0.1` | ❌ No match | ✅ Match | ✅ Valid |
| `127.0.0.1` | ❌ No match | ✅ Match | ✅ Valid |
| `999.999.999.999` | ✅ Match | ❌ No match | ❌ Invalid |
| `256.1.1.1` | ✅ Match | ❌ No match | ❌ Invalid |
| `0.0.0.0` | ❌ No match | ✅ Match | ✅ Valid |


k-rister:
@HughNhan Can you point me to the Python builtin you are referring to?

HughNhan:
> @HughNhan Can you point me to the Python builtin you are referring to?

import ipaddress

def is_valid_ip(ip_string):
    """Validate both IPv4 and IPv6 addresses"""
    try:
        ipaddress.ip_address(ip_string)
        return True
    except ValueError:
        return False

# Test examples
print(is_valid_ip("192.168.1.1"))      # True (IPv4)
print(is_valid_ip("192.168.0.1"))      # True (IPv4)
print(is_valid_ip("127.0.0.1"))        # True (IPv4)
print(is_valid_ip("0.0.0.0"))          # True (IPv4)
print(is_valid_ip("2001:db8::1"))      # True (IPv6)
print(is_valid_ip("::1"))              # True (IPv6)
print(is_valid_ip("999.999.999.999"))  # False
print(is_valid_ip("256.1.1.1"))        # False
print(is_valid_ip("not.an.ip"))        # False

PR #652, opened 2025-07-17
Ensure ssh_exception is imported before use in endpoints.py

natoscott:


k-rister (approved):

k-rister:
@natoscott We have some specific requirements for submitting PRs that require certain types of CI execution that I need to document somewhere.  I have invited you as a developer to our project which should allow you to do what I'm about to ask for once you've accepted.

Can you push your branch to the upstream repository and then open a PR from that branch into the master branch?  The PRs need to come from the upstream repository in order to have access to the secrets that are required to allow CI to run properly.

This used to not be the case so it is possible to run the jobs the way your PR did, but as you can see that doesn't work anymore.  In your case I went ahead and approved the submission just to see if it would work but it is pretty clear that at this point it does not so I need to figure out how to block that from ever happening since it's just a waste of time and resources.

@rafaelfolco ^^

PR #655, opened 2025-07-24
Fedora userenv updates

k-rister:


k-rister:
Having to fix a bug in the `get-userenvs` action: https://github.com/perftool-incubator/crucible-ci/pull/171

PR #659, opened 2025-07-30
Userenv updates

k-rister:


userenvs/alma10.json
```diff
@@ -0,0 +1,105 @@
...
+  "workshop": {
+    "schema": {
+      "version": "2024.08.07"
+    }
+  },
+  "userenv": {
+    "name": "alma10",
+    "label": "AlmaLinux 10",
+    "origin": {
+      "image": "docker.io/library/almalinux",
```
rafaelfolco, line 11: docker.io? shouldn't we change to quay.io?

rafaelfolco (changes requested):

userenvs/alma10.json
k-rister: Actually, it was originally quay.io (and it failed) and is on the previous Alma userenvs.  Apparently they have switched to using Docker Hub and that is the only registry where version 10 is located.  I assume the quay.io stuff is just used for backwards compatibility and going forward they will be exclusively using Docker Hub.

rafaelfolco (approved):

PR #661, opened 2025-08-05
Generate ci jobs

k-rister:


k-rister:
@rafaelfolco We should be able to merge this PR without potentially damanaging anything else because it's just adding files that nothing else should be referencing.  Then I can submit a PR to crucible-ci that will actually use these files and there we can experiment with making sure it works properly on a subset of the tests.

That said, feel free to point out any obvious problems (and request changes) and I'll resolve them before we merge this.

schema/crucible-ci.json
```diff
@@ -0,0 +1,169 @@
...
+                                                "tags",
+                                                "type"
+                                            ]
+                                        }
+                                    ]
+                                }
+                            },
+                            "additionalProperties": false,
+                            "required": [
+                                "enabled",
```
rafaelfolco, line 108: ok for being more explicit, this could be optional...

schema/crucible-ci.json
```diff
@@ -0,0 +1,169 @@
...
+                                    "denylist"
+                                ]
+                            }
+                        ]
+                    }
+                },
+                "additionalProperties": false,
+                "required": [
+                    "name",
+                    "enabled",
```
rafaelfolco, line 158: ditto

rafaelfolco (approved):
lgtm

schema/crucible-ci.json
k-rister: Yeah, originally I was going to make this optional and then I changed my mind and decided to make it required so that, if necessary, one could just go into the file and very quickly and easily flip the value from True to False or vice versa without having to worry about the schema and where the `enabled` property was allowed and not.  I'm not absolutely tied to that decision, that is just how I got to where it is currently at.

schema/crucible-ci.json
```diff
@@ -0,0 +1,169 @@
...
+                        "type": "boolean"
+                    },
+                    "scenarios": {
+                        "type": "array",
+                        "minItems": 1,
+                        "uniqueItems": true,
+                        "items": {
+                            "type": "object",
+                            "properties": {
+                                "enabled": {
```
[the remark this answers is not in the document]
k-rister: Yes, that was a good suggestion from you.

PR #670, opened 2025-08-12
Launch privileged pods for rtla on k8s

rafaelfolco:
remotehosts podman osruntime runs privileged by default


k-rister (approved):
Indentation crimes!

PR #689, opened 2025-08-25
update refresh expiration query to be resilient to errors in the retu…

k-rister:
…rned data

- if the returned data is invalid, such as an HTML page that says "Server Side Error" (or similar), the JSON decoder will throw an exception.  If the exception isn't handled, the Perl interpreter will die immediately.  So rework the logic to catch the exception and properly process the retry logic in the face of failures.


k-rister:
Seeing a bunch of CI failures, debugging.

k-rister:
CI failures were caused by https://github.com/perftool-incubator/crucible-ci/issues/176 and are being dealt with elsewhere.

PR #728, opened 2025-10-23
Shared engines dir

k-rister:


k-rister:
I should probably point out what I think some of the original reasons this shared directory was created are.

1. To enable cleanup of the SSH private key in the remotehosts endpoint.  In that situation, there are some files that are left over when a run is complete and even though the run's SSH private key should no longer be privileged to log into the controller we don't really want to leave it laying around.  By sharing this directory we only have a single location that need to clean up on each remote.
2. To share SSH multiplexed connections between engines.  This is done by using some specific SSH options ([see the current code](https://github.com/perftool-incubator/rickshaw/blob/3b8e3b2897e40678b18e003208254322cfc9fae7/engine/bootstrap#L51-L53)) to create a socket that multiple SSH command invocations can reuse to bypass going through the session creation, login, logout, and session teardown sequences over and over again.  This speeds things up substantially and in some situations (where many engines are running on a single remote) can significantly increase the scalability of the controller's SSH server since it reduces the total number of connections being made.

That is to say, the shared directory is for some very, very specific and nuanced things.

PR #855, opened 2026-07-15
feat: rescue engine logs and harden cleanup on early roadblock failure

### LINKED ISSUE #853: Engine abort failure counting bug causes catastrophic roadblock cascade on benchmark failure
## Summary

When a benchmark command fails (e.g., `trafficgen-infra` returns non-zero), the engine enters a catastrophic failure cascade: it starts a phantom second attempt that nobody else participates in, every remaining roadblock times out at 4 minutes each, and `wait_for_endpoints()` blocks indefinitely. A single benchmark failure produces 30-60+ minutes of wasted timeout followed by an infinite hang. Diagnostic files inside the engine container are destroyed before data collection can complete.

## Root cause

In `engine_lib.py` `_roadblock_and_evaluate()` (line 707), the guard condition for counting failures is:

```python
if result["is_abort"] and not self.abort:
```

When `run_bench_cmd()` detects a non-zero return code (line 850), it sets `self.abort = True` **before** `_roadblock_and_evaluate()` runs for the subsequent roadblock. The evaluate function sees `self.abort` is already True, so it **never increments the failure counter** and **never sets `sd["complete"] = True`**.

The while loop at line 801 then re-enters because:
- `not self.quit` → True
- `not self.abort` → True (reset to False at line 1082)
- `not sd["complete"]` → True (never set!)
- `sd["failures"] < max_sample_failures` → True (0 < 1, never incremented!)

The engine starts attempt 2, which nobody else participates in, causing timeouts that cascade through everything.

## Why controller and endpoint don't have this bug

The controller (`rickshaw-run.py`) uses a **local** `abort` variable (line 2134) that starts at 0 for each attempt. `evaluate_test_roadblock()` checks `if abort == 0` (the local var), so it correctly increments failures.

The endpoint (`endpoints.py`) similarly uses a local `abort` variable (line 1278) that starts `False`.

Only the engine uses `self.abort` (a persistent instance attribute) as both the "abort was signaled" flag AND the double-counting guard, creating the timing dependency.

## Verified impact

Tested with a trafficgen run where TRex fails to start on a ConnectX-6 LX NIC:

**Before fix:**
- All 16 bench roadblocks complete quickly via abort protocol (correct)
- Engine starts phantom attempt 2, times out on `1-1-2:infra-start-begin` (4 min)
- Post-bench roadblocks (stop-tools, send-data, endpoint-cleanup) each timeout (6 × 4 min = 24 min)
- `wait_for_endpoints()` blocks indefinitely
- Total: 30+ minutes wasted + infinite hang. Diagnostic files lost.

**After fix:**
- All 16 bench roadblocks complete quickly via abort protocol (correct)
- Engine correctly exits the while loop (failures=1 >= max_sample_failures=1)
- Post-bench roadblocks complete in **16 seconds total** (all participants synchronized)
- `send-data` completes successfully — **diagnostic files preserved**
- Run finishes cleanly with archived error output

## Fix

Change the guard from `not self.abort` to `not sample_data[iter_idx]["attempt-fail"]`. The `attempt-fail` flag is reset to False at the start of each attempt (line 806) and serves the same double-counting prevention purpose without the timing dependency.

k-rister:
## Summary

- When an engine fails during init-phase roadblocks (endpoint-deploy through start-tools), `process_roadblocks()` returns early and the `remote-cleanup` callback never fires — engine container logs are permanently lost. Add a `rescue-engine-logs` callback that fires via a `try/finally` guard on early exit, collecting engine logs without destroying containers or cleaning up the environment so the endpoint remains intact for debugging.
- `shutdown_engines_worker_thread` (remotehosts) and `kube_cleanup` (kube) can crash with an unhandled exception when `remote_connection()` fails after exhausting retries. In the worker thread case, this causes `work.join()` to hang indefinitely because `task_done()` is never called. Wrap remote operations in `try/except` so unreachable endpoints are logged as errors and execution continues gracefully.

## Design

- The `try/finally` in `process_roadblocks()` wraps only the roadblocks that have early returns (endpoint-deploy through process_bench_roadblocks), not the post-bench fire-and-forget roadblocks or the normal cleanup callback
- Rescue functions (`rescue_engine_logs`, `rescue_engine_logs_worker_thread` in remotehosts; `rescue_engine_logs` in kube) are separate from the normal cleanup path to allow independent evolution of failure-specific collection
- `collect_chroot_log()` gains a `remove` parameter so the rescue path preserves remote log files
- Rescue functions use WARNING log level for operational messages, ERROR at the entry point
- All rescue and cleanup functions are resilient to unreachable endpoints via `try/except` around remote connections

## Test plan

- [x] Success path: normal run completes with no rescue triggered
- [x] Failure path: injected `sys.exit(1)` after engine-init-begin — all 5 engine logs rescued (podman + chroot), containers left intact, no chroot log removal
- [x] Log levels: rescue messages at WARNING, entry point at ERROR, shared `collect_podman_log`/`collect_chroot_log` at their existing levels
- [x] Cleanup resilience: success path still works after adding `try/except` to `shutdown_engines_worker_thread` and `kube_cleanup`

Resolves: https://github.com/perftool-incubator/rickshaw/issues/853 (partial — improves error-path diagnostics)

### DIFF
--- a/endpoints/endpoints.py
+++ b/endpoints/endpoints.py
@@ -1021,6 +1021,7 @@ def process_roadblocks(callbacks = None, roadblock_id = None, endpoint_label = N
     else:
         logger.info("No new followers to inform the roadblock leader about")
 
+    # Separate flag because finally also fires on uncaught exceptions, not just rc != 0 returns
     roadblocks_completed = False
 
     try:
--- a/endpoints/endpoints.py
+++ b/endpoints/endpoints.py
@@ -983,6 +983,7 @@ def process_roadblocks(callbacks = None, roadblock_id = None, endpoint_label = N
                           test-start
                           test-stop
                           remote-cleanup
+                          rescue-engine-logs
         roadblock_id (str): The base ID to use as part of a roadblock's name
         endpoint_label (str): The name of the calling endpoint
         endpoint_deploy_timeout (int): The computed timeout value for endpoint engine deployment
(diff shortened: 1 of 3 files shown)


atheurer (commented):
Overall the PR is solid — the `try/finally` scoping is correct, the `task_done()` hang fix is sound, and the rescue-vs-cleanup separation is a reasonable design choice. Two actionable items before approval, plus one minor suggestion.

endpoints/kube/kube.py
```diff
@@ -2134,6 +2134,63 @@ def create_tools_pods(abort_event):
...
+    logger.warning("Rescuing engine logs")
+
+    endpoint = settings["run-file"]["endpoints"][args.endpoint_index]
+
+    if "pods" not in settings["engines"]["endpoint"] or not settings["engines"]["endpoint"]["pods"]:
+        logger.warning("No pods to rescue engine logs from")
+        return 0
+
+    try:
+        with endpoints.remote_connection(settings["run-file"]["endpoints"][args.endpoint_index]["host"],
```
atheurer: **Unused `endpoint` variable.** `endpoint` is assigned two lines above (line 2154) but `remote_connection()` indexes `settings` directly instead of using it. Should be:

```python
with endpoints.remote_connection(endpoint["host"], endpoint["user"]) as con:
```

(The same redundant long path is then used again inside the loop for `k8s_pod_name` on the line that falls back to `endpoint_default_settings["prefix"]["pod"]` — if you're cleaning up the above, worth checking if any other accesses in this function can use `endpoint` instead.)

endpoints/remotehosts/remotehosts.py
```diff
@@ -2267,6 +2272,101 @@ def image_mgmt():
...
+
+    while not work_queue.empty():
+        remote_idx = None
+        try:
+            remote_idx = work_queue.get(block = False)
+        except queue.Empty:
+            thread_logger("Received a work queue empty exception", log_level = "warning")
+            break
+
+        if remote_idx is None:
```
atheurer, line 2310: **Dead null-check without `task_done()`.** `None` is never enqueued — only integer indices go into `collect_work` — so this branch is unreachable. But if it were ever hit, the job would be dequeued without `task_done()` being called, and `work_queue.join()` would hang.

Either remove the check entirely (since `None` can't arrive here), or guard it properly:

```python
if remote_idx is None:
    thread_logger("Received a null job", log_level = "warning")
    work_queue.task_done()  # must ack even on bad input
    continue
```

Note: the same pattern exists in `shutdown_engines_worker_thread` (~line 2109) — whichever fix you pick here, apply it there too for consistency.

endpoints/endpoints.py
```diff
@@ -1020,130 +1021,144 @@ def process_roadblocks(callbacks = None, roadblock_id = None, endpoint_label = N
...
-    rc = do_roadblock(roadblock_id = roadblock_id,
-                      follower_id = endpoint_label,
-                      label = "start-tools-end",
-                      timeout = roadblock_timeouts["default"],
-                      redis_password = roadblock_password,
-                      msgs_dir = roadblock_messages_dir,
-                      connection_watchdog = roadblock_connection_watchdog)
-    if rc != 0:
-        return rc
+    roadblocks_completed = False
```
atheurer, line 1025: **Minor suggestion.** The purpose of this flag isn't immediately obvious — `rc` is already tracked, so a reader may wonder why a separate boolean is needed. A one-liner would help:

```python
# Separate flag because finally also fires on uncaught exceptions, not just rc != 0 returns
roadblocks_completed = False
```

atheurer (approved):
### 📝 PR Review (Approved)
_Reviewed in `agy` using the **Gemini 3.5 Flash (High)** model._

---

#### 🔍 Overall Summary
This is an exceptionally solid and well-engineered pull request. The introduction of the `rescue-engine-logs` callback via the `try/finally` guard in `process_roadblocks()` effectively addresses a critical diagnostics gap where logs were permanently lost on early roadblock exits. Additionally, the hardening of `shutdown_engines_worker_thread` and `kube_cleanup()` against unreachable endpoints makes the tear-down phase highly resilient.

The follow-up commit `35c8bb7` ("address review feedback for rescue-engine-logs PR") completely resolves all previous code-quality suggestions, ensuring thread queues never hang and local variables are properly utilized.

---

#### 🛠️ Key Architectural Strengths

1. **Robust Try/Finally Scoping in `process_roadblocks()`**
   - The `roadblocks_completed` boolean flag safely handles early exit scenarios (whether due to a return code non-zero or an uncaught exception) without masking any original errors.
   - Restricting the guard scope to init-phase roadblocks (up to and including `process_bench_roadblocks`) is the correct architectural choice to prevent the rescue logic from interfering with normal post-test collection.

2. **Log Preservation via `remove = False`**
   - Adding the `remove` parameter to `collect_chroot_log` (defaulting to `True` but overridden to `False` in the rescue path) ensures that remote log files remain fully intact on the endpoint during early failure scenarios, preserving vital state for active troubleshooting.

3. **Multi-Threaded Queue Safety (Addressing Thread Hangs)**
   - Incorporating `work_queue.task_done()` inside the `None` (null job) guards across all worker threads (such as `image_pull`, `remote_mkdirs`, `launch_engines`, `shutdown_engines`, `collect_sysinfo`, and `rescue_engine_logs`) guarantees that worker pool `join()` operations will never hang on toxic/unmatched queue inputs.

4. **Excellent DRY Refactoring**
   - Using the local `endpoint` variable in Kubernetes `rescue_engine_logs()` rather than repeatedly indexing the nested `settings` structure significantly improves code readability and maintainability.

---

#### 📋 Detailed Walkthrough of Changes

##### 1. `endpoints/endpoints.py`
* Wrap roadblock execution sequence inside `try/finally` blocks, using the `roadblocks_completed` state flag to conditionally trigger `rescue-engine-logs`.
* Highly robust; any exception during roadblock callbacks is captured, logged, and prevented from masking original roadblock execution errors.

##### 2. `endpoints/kube/kube.py`
* Introduced the namespace-safe `rescue_engine_logs()` function, wrapped `kube_cleanup()` in generic exception guards, and registered the new callback in `main()`.
* Safe and clean. By utilizing the locally-assigned `endpoint` variable for remote operations, readability is significantly improved.

##### 3. `endpoints/remotehosts/remotehosts.py`
* Defined the multi-threaded `rescue_engine_logs()` mechanism using a dedicated thread pool and the new worker thread `rescue_engine_logs_worker_thread`. Applied `work_queue.task_done()` fixes to all worker threads.
* Perfect thread queue discipline. The concurrency model is safe, clean, and avoids potential deadlocks.

---

#### 🌟 Verdict
This is an outstanding contribution that dramatically improves Rickshaw's failure-path diagnostic capabilities. **LGTM!**

atheurer:
## Code Review — Claude Code (Opus 4.6)

**Overall: This is a solid PR that fixes two real problems.** The design is sound and follows existing codebase patterns.

### What it does

1. **Rescues engine logs on early roadblock failure** — When init-phase roadblocks fail (endpoint-deploy through start-tools), `process_roadblocks()` returns early and the `remote-cleanup` callback never fires, losing engine container logs permanently. The fix adds a `try/finally` guard that calls a new `rescue-engine-logs` callback on early exit.

2. **Fixes hung worker threads** — `shutdown_engines_worker_thread` and `kube_cleanup` could crash on unreachable endpoints, and since `task_done()` was never called, `work.join()` would hang indefinitely. Now wrapped in `try/except`.

3. **Fixes pre-existing null-job `task_done()` bugs** — Six worker threads had a path where a `None` job causes `continue` without calling `work_queue.task_done()`, which would also hang the join. These are pre-existing bugs, not introduced by this PR.

### What looks good

- **The `try/finally` pattern in `process_roadblocks()`** is correct. The `roadblocks_completed` flag is the right approach — it correctly distinguishes "completed normally" from both non-zero-rc early returns and uncaught exceptions. The rescue callback is wrapped in its own `try/except` so it can't mask the original error or alter the return value.

- **Rescue vs cleanup separation** is a good design choice. The rescue path deliberately doesn't destroy containers or clean up the environment, leaving the endpoint intact for debugging. The `remove=False` parameter on `collect_chroot_log` preserves remote log files for the same reason.

- **The worker thread `task_done()` fixes** are important correctness fixes. Every `continue` path must call `task_done()` or the queue join hangs forever.

### Concerns

1. **Rescue log directory may not exist yet.** If the failure happens very early (before `settings["dirs"]["local"]["engine-logs"]` is created), `lzma.open()` will raise. This is caught by the outer `try/except`, so it won't crash, but you'll get a generic "Failed to rescue engine logs" error with no logs saved. A `makedirs(exist_ok=True)` before writing would make the rescue path more resilient. Worth a follow-up or defensive add.

2. **`kube_cleanup` now returns 0 on connection failure.** Previously it would propagate the exception. The PR description says this matches the convention (cleanup functions return 0, communicate errors via logging), which seems accurate based on the rest of the code. But double-check whether any caller actually checks this return code — if so, swallowing the exception changes behavior.

3. **Minor: code duplication in kube rescue vs cleanup.** The pod iteration + `kubectl logs` + lzma compression logic is duplicated between `rescue_engine_logs()` and `kube_cleanup()`. The PR description calls this out as intentional for independent evolution. That's reasonable, but if the duplication grows, a shared helper with a `destroy` flag (similar to `collect_chroot_log`'s `remove` flag) would be cleaner.

4. **Typo carried forward**: `"Deleting namepsace"` (misspelling of "namespace") was in the original code and persists in `kube_cleanup`. Not introduced by this PR but easy to fix while you're here.

### Verdict

Good to merge. The core logic is correct, the design is well-motivated, and the null-job `task_done()` fixes alone are worth the PR. The log directory concern (#1) is the only one I'd consider addressing before merge — everything else is minor or follow-up material.

---
*Review by Claude Code (claude-opus-4-6, 1M context)*

PR #858, opened 2026-07-20
feat: support multi-stage workshop files for benchmarks and tools

k-rister:
## Summary

- Allow benchmarks and tools to split a workshop file into ordered sub-stages using numbered files (`workshop-NN-name.json` or `{role}-workshop-NN-name.json`), where each file becomes its own cached build stage
- Enables independent cache invalidation — changing one component (e.g., ptp-latency) no longer forces rebuilding unrelated components (e.g., TRex) in the same benchmark
- Fully backwards compatible — plain `workshop.json` / `client-workshop.json` / `server-workshop.json` continue to work unchanged

## Test plan

- [x] Run a benchmark with a single `workshop.json` (e.g., fio) — should work unchanged
- [x] Run trafficgen with split `client-workshop-01-trex.json` + `client-workshop-02-ptp-latency.json` — should produce two client stages
- [x] Change only `ptp-latency.c`, run again — only the 02-ptp-latency stage should rebuild; 01-trex should be a cache hit
- [x] Verify stage descriptions in SIS log output show descriptive labels

### DIFF
--- a/rickshaw-run.py
+++ b/rickshaw-run.py
@@ -6,6 +6,7 @@
 # endpoints, sources container images, deploys engines via roadblock
 # synchronization, and organizes results.
 
+import glob
 import json
 import os
 import platform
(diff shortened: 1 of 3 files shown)


atheurer (changes requested):
Thank you for this elegant and well-thought-out implementation! Splitting a `workshop.json` file into ordered sub-stages is a great feature that will greatly benefit incremental builds and cache reuse.

The design choice of remapping split workshop filenames client-side so that the `source-images-service` remains completely role-agnostic is highly elegant and works seamlessly.

During detailed review of the changes, we identified two main findings (one critical bug and one sorting limitation/guideline) that should be addressed before merging.

---

### 1. Critical Bug: Silent File Exclusion & Incorrect Renaming

#### The Issue
In `rickshaw-source-images-client.py`, the `_remap_workshop_for_role` function uses simple prefix checks to identify and exclude or rename workshop files:

```python
    for entry in collected["files"]:
        fname = entry["filename"]
        if fname.startswith(other_prefixes) or fname.startswith("workshop"):
            continue
        if fname.startswith(f"{role_prefix}workshop"):
            remapped_files.append({
                "filename": fname[len(role_prefix):],
                ...
```

If a benchmark directory contains non-workshop files at its root whose names happen to start with `"workshop"`, `"client-workshop"`, or `"server-workshop"`, this logic will match them. For example:
- **`workshop-helper.sh` or `workshop_readme.md`**: Since these start with `"workshop"`, they will match `fname.startswith("workshop")` and be silently excluded from the collected files sent to the build server.
- **`client-workshop-utils.py`** (when role is `client`): Since this starts with `"client-workshop"`, it will match `fname.startswith(f"{role_prefix}workshop")` and be renamed to `"workshop-utils.py"`.

#### The Fix
To ensure we only exclude or rename actual workshop JSON files, we should restrict the check to files that end with `.json`. Here is the proposed correction:

```python
def _remap_workshop_for_role(collected, role):
    """Remap a collected directory for a role-specific workshop build.

    For split workshop benchmarks, rename {role}-workshop*.json files to
    workshop*.json (stripping the role prefix) and exclude the other role's
    workshop files so the source-images-service sees standard workshop*.json.

    Supports both single-file ({role}-workshop.json -> workshop.json) and
    multi-stage ({role}-workshop-NN-name.json -> workshop-NN-name.json).
    """
    role_prefix = f"{role}-"
    other_roles = {"client", "server"} - {role}
    other_prefixes = tuple(f"{r}-workshop" for r in other_roles)

    remapped_files = []
    for entry in collected["files"]:
        fname = entry["filename"]
        
        # Identify actual workshop JSON files
        is_workshop_json = fname.endswith(".json") and (
            fname.startswith("workshop") or
            fname.startswith("client-workshop") or
            fname.startswith("server-workshop")
        )
        
        if is_workshop_json:
            if fname.startswith(other_prefixes) or fname.startswith("workshop"):
                continue
            if fname.startswith(f"{role_prefix}workshop"):
                remapped_files.append({
                    "filename": fname[len(role_prefix):],
                    "content_base64": entry["content_base64"],
                    "mode": entry["mode"],
                })
        else:
            remapped_files.append(entry)

    remapped_files.sort(key=lambda f: f["filename"])
    return {"name": collected["name"], "files": remapped_files}
```

---

### 2. Multi-Stage Sorting Limitation (Padded Numbers Requirement)

#### The Issue
In `requirements_builder.py`, the stages are sorted using Python's standard `sorted()` function:

```python
    workshop_files = sorted(bench_dir.glob("workshop*.json"))
```

Because `sorted()` on strings performs lexicographical (alphabetical) sorting, numeric stage values that are not zero-padded will sort out of numerical order. For example:
- `workshop-1.json`, `workshop-2.json`, ..., `workshop-10.json`
- Sorted order: `workshop-1.json`, `workshop-10.json`, `workshop-2.json`

This will cause `workshop-10.json` to be built as Stage 2, before `workshop-2.json`, which will break builds that expect sequential progression.

#### The Recommendation
There are two ways to address this:
1. **User Documentation / Guidelines (Simple)**: Explicitly document that all multi-stage workshop filenames must use zero-padded numeric suffixes (such as `workshop-01-fio.json`, `workshop-02-other.json`, `workshop-10-etc.json`) to guarantee correct sorting.
2. **Natural Sorting (Robust)**: Use a natural sorting key when sorting the files. For example, using Python's `re` module:

```python
import re

def natural_sort_key(path):
    return [int(text) if text.isdigit() else text.lower() for text in re.split(r'(\d+)', str(path))]

...
workshop_files = sorted(bench_dir.glob("workshop*.json"), key=natural_sort_key)
```

---

### Notable Merits

* **Role-Agnostic Build Backend**: Remapping role-specific prefixes client-side (`client-workshop-*.json` -> `workshop-*.json`) before transmitting the directory package is extremely clean.
* **Seamless Tool Support**: Because tools are registered in `self.bench_dirs` and mapped during the build process, they automatically benefit from the multi-stage workshop stage building logic without requiring separate changes.
* **Robust Configuration Validation**: The checks in `rickshaw-run.py` to prevent coexistence of plain workshops and split/numbered workshops are well-designed and prevent conflicting definitions.

atheurer (approved):
All concerns from the previous review have been successfully addressed. The file remapping now robustly handles non-JSON files, and natural sorting has been added for stage files. Excellent work! Approving.

PR #862, opened 2026-08-04
feat: validate rickshaw.json against schemas at load time

k-rister:
## Summary
- Add `validate_schema()` calls for benchmark, tool, and utility `rickshaw.json` files at load time in `rickshaw-run.py`
- Make `controller` optional in the tool schema — tools like ftrace and kernel legitimately have no post-processing, and the code already handles this gracefully
- The schema files and validation infrastructure already existed but were never connected; this change closes that gap

Addresses [PERFNFV-454](https://redhat.atlassian.net/browse/PERFNFV-454) from the architecture review finding that rickshaw.json schemas were defined but never applied.

## Test plan
- [ ] Verified all 16 benchmarks and 13 tools pass schema validation (including ftrace and kernel which have no controller section)
- [ ] Verified invalid input is correctly rejected (typos like `fils-from-controller`, missing required fields like `start`)
- [ ] Integration test with `crucible run` using a known-good run file

### DIFF
--- a/rickshaw-run.py
+++ b/rickshaw-run.py
@@ -30,7 +30,7 @@ if RICKSHAW_HOME:
     sys.path.append(RICKSHAW_HOME)
 
 from toolbox.fileio import open_write_text_file
-from toolbox.json import load_json_file, save_json_file
+from toolbox.json import load_json_file, save_json_file, validate_schema
 from toolbox.jsonsettings import get_json_setting
 import logging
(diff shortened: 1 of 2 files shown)


atheurer (approved):
Approve. Closes a real gap between existing schema infrastructure and actual validation (validate_schema() was previously dead code); the tool.json required-field fix matches actual code behavior for tools without post-processing. CI green across the full 206-check matrix.

